PR #33, opened 2026-03-16
Create and/or document a demo setup for running the flow end-to-end

Ogenbertrand:
Closes: https://github.com/ADORSYS-GIS/digital-identity-arch-review-mab/issues/3


Vitalisn4 (commented):
####  Demo Failure  as `"vc+sd-jwt"` is not Recognized

Running the demo locally (` just -f demo/Justfile demo or ./demo/run-demo.sh`) fails at **Step 2** when parsing the signed request object in `sample-flow.java` (line 79):

```
UnrecognizedPropertyException: Unrecognized field "vc+sd-jwt"
(class ClientMetadata$VpFormat) not marked as ignorable
```

#### Root Cause

The plugin emits `"vc+sd-jwt"` in `client_metadata.vp_formats`, but `ClaimFormat.java` only defines `"dc+sd-jwt"`.
Since the class does not ignore unknown fields, Jackson throws an exception.

#### Possible Fixes

One of the following may resolve it:

1. **Add `"vc+sd-jwt"`** to `ClaimFormat` if it is a valid format.
2. **Add `@JsonIgnoreProperties(ignoreUnknown = true)`** to tolerate unknown fields.
3. **Update the server to emit `"dc+sd-jwt"`** if `"vc+sd-jwt"` is incorrect.

Vitalisn4 (dismissed):
LGTM

IngridPuppet (changes requested):

IngridPuppet (dismissed):
I tested and confirm that the demo works as expected.

TBH however, I'm overwhelmed by the absence of comments and the lack of structure in the generated code. I could not review it line by line and do not consider it (human-)maintainable.

Since it's just a demo, I think it is okay to merge it until new concerns arise.

IngridPuppet (approved):
LGTM. Thank you for the update.

PR #39, opened 2026-03-18
 Do not leak detailed reasons in error responses

Vitalisn4:
#### Summary

- Return generic client-facing error_description messages by default and keep detailed verification causes in server logs only.
- Add correlation IDs (ref: <uuid>) to client errors for support/debugging.
- Introduce OID4VP_VERBOSE_ERRORS=true / -Doid4vp.verboseErrors=true to re-enable verbose error descriptions when required.
- Ensure status polling (/status/{transactionId}) does not expose detailed internal failure reasons.

#### Test plan

docker ps
./mvnw test

(Optional) verify verbose mode: OID4VP_VERBOSE_ERRORS=true ./mvnw test or run Keycloak with -Doid4vp.verboseErrors=true and confirm detailed errors return.

Closes https://github.com/ADORSYS-GIS/digital-identity-arch-review-mab/issues/7

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,25 +69,12 @@ you are encouraged to create a `docker-compose.override.yml` based on the provid
 
 ## Configuration
 
-This plugin supports a `verbose-errors` setting to control whether detailed verification errors are returned to
-clients. By default, verbose errors are disabled (safe-by-default).
+This plugin supports a `verbose-errors` setting to control whether detailed verification errors are returned to clients.
+By default, verbose errors are disabled (safe-by-default).
 
-Provider configuration should use Keycloak's SPI configuration mechanism (e.g., `keycloak.conf`, `KC_*` environment
-variables, or CLI flags), which is exposed to providers via `init(Config.Scope config)`.
-
-Examples (from Keycloak docs: `spi-<spi-id>--<provider-id>--<property>=<value>`):
-
-- Enable verbose errors for the OpenID4VP user auth endpoint provider:
-
-  `spi-realm-restapi-extension--oid4vp-auth--verbose-errors=true`
-
-- Enable verbose errors for the login actions provider:
-
-  `spi-realm-restapi-extension--oid4vp-login-actions--verbose-errors=true`
-
-- Enable verbose errors for the SD-JWT authenticator provider:
-
-  `spi-authenticator--sd-jwt-authenticator--verbose-errors=true`
+For provider configuration (including examples for `keycloak.conf`, `KC_*` environment variables, and CLI flags), see the
+documentation site sources in `docs/`, especially
+[`docs/modules/ROOT/pages/non-functional-requirements.adoc`](./docs/modules/ROOT/pages/non-functional-requirements.adoc).
 
 ## Documentation site (Antora)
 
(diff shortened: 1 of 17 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPConfig.java
```diff
@@ -0,0 +1,49 @@
...
+    private OID4VPConfig() {}
+
+    public static boolean verboseErrors() {
+        return verboseErrors;
+    }
+
+    public static synchronized void init(Config.Scope config) {
+        if (initialized) {
+            return;
+        }
```
Ogenbertrand: ```init()``` is called from multiple factoriesz, the first call wins and later scopes are ignored. If verboseErrors is only configured in a later provider scope, it will be dropped depending on init order.

Ogenbertrand (changes requested):
Thanks for this PR @Vitalisn4 !
I added a few comments please check them.

IngridPuppet (changes requested):
I left some initial comments but will need to check the changes in-depth in my next review iteration. I am not 100% okay with the ErrorResponseSanitizer construct introduced.

README.md
```diff
@@ -71,24 +64,30 @@ docker compose up
...
+
+  `spi-realm-restapi-extension--oid4vp-auth--verbose-errors=true`
+
+- Enable verbose errors for the login actions provider:
+
+  `spi-realm-restapi-extension--oid4vp-login-actions--verbose-errors=true`
+
+- Enable verbose errors for the SD-JWT authenticator provider:
+
+  `spi-authenticator--sd-jwt-authenticator--verbose-errors=true`
```
IngridPuppet: Please look up where the other config is documented. We would like to keep this README short.

IngridPuppet (changes requested):
I added more comments. Please could you check?

IngridPuppet (changes requested):
Great update! I added a few more comments. Please could you check?

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointTest.java
```diff
@@ -336,8 +336,8 @@ public void shouldFailAuthentication_InvalidSdJwtVPToken_Empty() throws Exceptio
 
         // Assert error response
         OAuth2ErrorRepresentation errorRep = parseErrorResponse(response);
-        assertEquals(
-                "Unparseable response params (vp_token must not be null or blank)", errorRep.getErrorDescription());
+        assertEquals(OAuthErrorException.INVALID_REQUEST, errorRep.getError());
+        assertTrue(errorRep.getErrorDescription().contains("Unparseable response params"));
```
IngridPuppet:
```suggestion
        assertTrue(errorRep.getErrorDescription().contains("Unparseable response params (vp_token must not be null or blank)"));
```

IngridPuppet (commented):
One last comment

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointTest.java
IngridPuppet: Please bring back the assertion of the error reason. It is important for test reliability.

IngridPuppet (approved):
LGTM. Thank you for the update.

Ogenbertrand (approved):
Thanks for adressing my comements @Vitalisn4 !
PR LGME

PR #40, opened 2026-03-19
Account Binding Hardening

Ogenbertrand:
PR Summary (Account Binding Hardening)

- Switched SD-JWT binding from `username` to stable `sub` and resolve users via `users().getUserById(realm, sub)`.
- Added configurable issuer validation (`verifyIssuerClaim`, default true). When disabled, issuer mismatches are logged as warnings (not fatal).
- Updated presentation requirements to require `sub` disclosure.
- Refreshed tests and realm imports to include a stable test user ID and ensure test SD-JWTs carry `sub`.
- Added debug logs for subject resolution and issuer checks (no auth codes or secrets logged).

Tests: `./mvnw -Dtest=OID4VPUserAuthEndpointTest test`


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthenticator.java
```diff
@@ -144,6 +168,39 @@ private JsonNode readSelectivelyDisclosedUsername(SdJwtVP sdJwt) {
...
+
+        if (!expectedIssuer.equals(issuer)) {
+            logger.warnf("Issuer mismatch. Expected '%s' but got '%s'", expectedIssuer, issuer);
+            failRejectingPresentedSdJwtToken(context, "Issuer mismatch");
+            return false;
+        }
+        logger.debugf("Issuer claim verified: %s", issuer);
+
+        return true;
+    }
```
IngridPuppet: I was quite certain that this  was what you had in mind. The logic was there in past, but later removed because it causes so many issues in practice without real security benefits. Verifying the signature of the credential with trusted keys is a stronger assurance of its origin than checking this issuer claim. Also notice that this is not used to identify the authenticating user. Only the sub claim is used for binding, which is fine IMO.

However, you came up with the idea of making this claim configurable. I totally buy this idea. But there is a very simple alternative to this long method. Please check this old code and adjust. `keycloakIssuerURI` is instantiated in the constructor of SdJwtAuthRequirements.

https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/blob/9d4dd5fa715bac697877f079cd61403c0a53f3a2/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthRequirements.java#L97-L109

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthRequirements.java
```diff
@@ -78,15 +82,18 @@ public List<String> getExpectedVcts() {
     }
 
     public List<String> getRequiredClaims() {
-        // A username field is required so as to reliably recover
-        // the user associated with the presented credential
-        return List.of(OAuth2Constants.USERNAME);
+        // A subject is required so we can recover the user by stable identifier
+        return List.of(JsonWebToken.SUBJECT);
     }
```
IngridPuppet: Please require both otherwise many wallets will not present the username. Yet, you're checking it as well.

IngridPuppet (changes requested):
Hi @Ogenbertrand - As per my review, the PR implements a very valid solution to the issue with the expected effect of hardening account binding. The implementation is also close to perfect in my opinion. I left some comments for your consideration. Please could you check?

IngridPuppet (approved):
LGTM

src/test/resources/realms/test-realm.json
```diff
@@ -18,6 +19,7 @@
       ]
     },
     {
+      "id": "disabled-user-id",
       "username": "disabled-user@localhost",
```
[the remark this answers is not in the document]
IngridPuppet: I'll do this in my PR.

PR #43, opened 2026-04-08
Reject non oid4vp codes in wallet login form

IngridPuppet:
Closes https://github.com/ADORSYS-GIS/digital-identity-arch-review-mab/issues/13

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/OID4VPLoginActionsService.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/OID4VPLoginActionsService.java
@@ -1,8 +1,10 @@
 package io.github.adorsysgis.keycloak.protocol.oid4vc.oidc;
 
 import static io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.service.AuthorizationResponseService.PARENT_AUTH_SESSION_ID;
-import static io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.service.AuthorizationResponseService.SCOPE_OPENID4VP;
+import static io.github.adorsysgis.keycloak.protocol.oid4vc.oidc.freemarker.OID4VPUserAuthBean.LOGIN_METHOD_OID4VP;
+import static io.github.adorsysgis.keycloak.protocol.oid4vc.oidc.freemarker.OID4VPUserAuthBean.PARAM_LOGIN_METHOD;
 
+import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.OID4VPUserAuthEndpointBase;
 import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.FormParam;
 import jakarta.ws.rs.POST;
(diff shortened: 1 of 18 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/OID4VPLoginActionsService.java
```diff
@@ -110,9 +118,19 @@ public Response oid4vpAuthLogin(
         logger.debug("Validating authorization code");
         OAuth2CodeParser.ParseResult result = OAuth2CodeParser.parseCode(session, authorizationCode, realm, event);
```
Ogenbertrand, line 119: `OAuth2CodeParser.parseCode()` removes the single-use code from Keycloak's store before the OID4VP-specific checks below run, so a wrong-session submission can burn a valid wallet-issued code before the rightful browser tab uses it. Please cover the rejected-then-correct retry path in a regression test, and move the binding check somewhere that does not consume the code first.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPLoginActionsServiceTest.java
```diff
@@ -0,0 +1,137 @@
...
+        String sdJwt = sdJwtVPTestUtils.requestSdJwtCredential(VCT_CONFIG_DEFAULT, TEST_USER);
+
+        // Complete authentication with alternate context
+        TestOpts opts =
+                TestOpts.getDefault().setAuthorizationContext(authContext).setShouldRetrieveAccessToken(false);
+        String authCode = testSuccessfulAuthentication(sdJwt, opts);
+        shouldFailAuthenticationWithAltAuthCode(authCode, "Authorization code was not issued for this OIDC session");
+    }
+
+    private void shouldFailAuthenticationWithAltAuthCode(String authCode, String reason) throws Exception {
```
Ogenbertrand, line 118: This helper  never retries the same valid OID4VP code in the rightful browser session after that rejection. That missing assertion is important because `parseCode()` consumes the code before the new context checks run.

Ogenbertrand (commented):
Thank you @IngridPuppet  for this PR!
I left some monor observation, could check and give me your feedback ?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/OID4VPLoginActionsService.java
IngridPuppet: I acknowledge this is not something I realized when making these changes. But thinking of it, why should we bother enforcing that a stolen code remains valid? It's actually a good thing in my opinion that it is quickly invalidated on first illegal use. The rightful user could still retry logging in.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPLoginActionsServiceTest.java
IngridPuppet: Same point as above. It's actually a good thing that the stolen code cannot be reused.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/OID4VPLoginActionsService.java
Ogenbertrand: True, but I'd still add a test to document that this is intentional rather than an incidental side effect of the current validation order.

IngridPuppet: Done.

Ogenbertrand (approved):
Thanks for considering my comments @IngridPuppet !
I have tested the changes and confirm the PR addresses the concern.

IngridPuppet:
Thank you for the review @Ogenbertrand !

PR #46, opened 2026-04-10
Temporary username fallback in SD‑JWT user binding

Ogenbertrand:
Adds a short‑term fallback in recoverAuthenticatingUser to resolve users by username when the subject ID doesn’t map, with a clear TODO to remove once the SubjectID mapper is fixed. This unblocks current flows while keeping username checks intact.


IngridPuppet (changes requested):
Please add a test with a valid username and invalid ID. The test must pass.

Ogenbertrand:
Done @IngridPuppet

IngridPuppet (approved):
LGTM. Please squash when merging. Then create next patch tag to trigger release.

Awambeng (approved):
Nitpick: It might be helpful to add more logging in failure paths to make troubleshooting easier in case something goes wrong.

PR #47, opened 2026-04-11
Advertise schemed client IDs by default

IngridPuppet:
Schemed client IDs should be used in more places, beyond the authorization request link.

https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-syntax
> In the client_id Authorization Request parameter and other places where the Client Identifier is used, the Client Identifier Prefixes are prefixed to the usual Client Identifier

Check examples of advertised request objects
https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-examples

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ClientMetadata.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ClientMetadata.java
@@ -1,8 +1,10 @@
 package io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ClientMetadata.java
```diff
@@ -29,6 +31,13 @@ public ClientMetadata setClientId(String clientId) {
         return this;
     }
 
+    @JsonIgnore
+    public String getSchemelessClientId() {
+        return Optional.ofNullable(clientId)
+                .map(id -> id.substring(id.indexOf(':') + 1))
```
Awambeng, line 37: It might be worth adding validation for the format just in case the `:` isn't present.

Awambeng (changes requested):

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ClientMetadata.java
IngridPuppet: `indexOf` returns -1 when ':' is not present. That means the whole string will be returned, which is the correct behavior IMO.

Awambeng (approved):

Issue #49, opened 2026-04-14
Add support for externally signed certificates


IngridPuppet:
Currently, authorization requests embed a self-signed certificate that may not be trusted by other parties. Configure plugin to work with externally signed certificates.

ETA: 2 days

IngridPuppet:
```json
  "verifier_info":{
    "data": "eY43rfc...",
    "format": "registration_cert"
  },
```

```json
  "response_mode": "direct_post.jwt",
  "scope": "openid",
  "nonce": "n-abcdef1234567890",
  "state": "s-xyz987654321",
  "redirect_uri": "https://berlin.digitalbank.example/post-registration",
  "client_metadata": {
    "jwks": {
      "keys": [
        {
          "kty": "EC",
          "kid": "ec-key-1",
          "use": "enc",
          "alg": "ES256",
          "crv": "P-256",
          "x": "f83OJ3D2xF4yG9mXQnKkT7H5Y3WZxR5m7yZy2P1XQ0",
          "y": "x_FEzRu9zM6T5nHcJZJx4Z3VtP4yKQz5xA6cD7E8F9"
        }
      ]
    }
  },
```

Support same device flow with cookieId and redirect uri: https://bmi.usercontent.opencode.de/eudi-wallet/wallet-development-documentation-public/latest/architecture-concept/flows/22-pid-presentation/#sequence-diagram

IngridPuppet:
First PR sent for review:
- https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/52

Next, add support for same-device flow as per [PID Presentation](https://bmi.usercontent.opencode.de/eudi-wallet/wallet-development-documentation-public/latest/architecture-concept/flows/22-pid-presentation/).

IngridPuppet:
Progress for same-device flow:
https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/tree/49-support-same-device-flow

PR #52, opened 2026-04-15
Add support for passing externally-signed certificates and response encryption

IngridPuppet:
This PR addresses two main points:
- Configuration for certificates to advertise and other minor details (auth request URL scheme, response mode to enforce)
- Response encryption/decryption with ephemeral keys

These are necessary to achieve conformance with [PID Presentation](https://bmi.usercontent.opencode.de/eudi-wallet/wallet-development-documentation-public/latest/architecture-concept/flows/22-pid-presentation/).

Related to #49

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/VerifierConfig.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/VerifierConfig.java
@@ -58,11 +58,11 @@ public class VerifierConfig {
 
     private static ResponseMode validateResponseMode(String responseMode) {
         try {
-            return ResponseMode.valueOf(responseMode);
+            return ResponseMode.fromValue(responseMode);
         } catch (IllegalArgumentException e) {
             String defaultResponseMode = SdJwtAuthenticatorFactory.RESPONSE_MODE_CONFIG_DEFAULT;
             logger.warnf("Invalid response mode: %s. Defaulting to %s", responseMode, defaultResponseMode);
-            return ResponseMode.valueOf(defaultResponseMode);
+            return ResponseMode.fromValue(defaultResponseMode);
         }
     }
 
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/VerifierConfig.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/VerifierConfig.java
@@ -1,0 +1,119 @@
+package io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.config;
+
+import com.apicatalog.jsonld.StringUtils;
+import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.authenticator.SdJwtAuthRequirements;
+import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.authenticator.SdJwtAuthenticatorFactory;
+import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model.ResponseMode;
+import java.io.ByteArrayInputStream;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.util.Base64;
+import java.util.Map;
+import org.jboss.logging.Logger;
+import org.keycloak.models.AuthenticatorConfigModel;
+import org.keycloak.models.KeycloakContext;
+
+/**
+ * Access configurations that modulate the verifier's behavior.
+ * <p></p>
+ * Read full descriptions of configurations in {@link SdJwtAuthenticatorFactory}.
+ */
+public class VerifierConfig {
+
+    private static final Logger logger = Logger.getLogger(VerifierConfig.class);
+
+    private final SdJwtAuthRequirements authRequirements;
+
+    private final ResponseMode responseMode;
+    private final String authReqUrlScheme;
+    private final X509Certificate accessCertificate;
+    private final String registrationCertificate;
+
+    public VerifierConfig(KeycloakContext context, AuthenticatorConfigModel authConfig) {
+        logger.debugf("Collecting verifier config properties");
+
+        Map<String, String> config =
+                (authConfig != null && authConfig.getConfig() != null) ? authConfig.getConfig() : Map.of();
+
+        // TODO: Relocate these non-SD-JWT-specific configurations.
+        //  They should normally not be exposed through SdJwtAuthenticatorFactory.
+
+        this.responseMode = validateResponseMode(config.getOrDefault(
+                SdJwtAuthenticatorFactory.RESPONSE_MODE_CONFIG,
+                SdJwtAuthenticatorFactory.RESPONSE_MODE_CONFIG_DEFAULT));
+
+        this.authReqUrlScheme = validateCustomUrlScheme(config.getOrDefault(
+                SdJwtAuthenticatorFactory.CUSTOM_URL_SCHEME_CONFIG,
+                SdJwtAuthenticatorFactory.CUSTOM_URL_SCHEME_CONFIG_DEFAULT));
+
+        this.accessCertificate =
+                validateX5CCertificate(config.get(SdJwtAuthenticatorFactory.ACCESS_CERTIFICATE_CONFIG));
+
+        this.registrationCertificate = config.get(SdJwtAuthenticatorFactory.REGISTRATION_CERTIFICATE_CONFIG);
+
+        // Collect authentication requirements
+        this.authRequirements = new SdJwtAuthRequirements(context, authConfig);
+    }
+
+    private static ResponseMode validateResponseMode(String responseMode) {
+        try {
+            return ResponseMode.valueOf(responseMode);
+        } catch (IllegalArgumentException e) {
+            String defaultResponseMode = SdJwtAuthenticatorFactory.RESPONSE_MODE_CONFIG_DEFAULT;
+            logger.warnf("Invalid response mode: %s. Defaulting to %s", responseMode, defaultResponseMode);
+            return ResponseMode.valueOf(defaultResponseMode);
+        }
+    }
+
+    private static String validateCustomUrlScheme(String customUrlScheme) {
+        String defaultCustomUrlScheme = SdJwtAuthenticatorFactory.RESPONSE_MODE_CONFIG_DEFAULT;
+        if (StringUtils.isBlank(customUrlScheme)) {
+            return defaultCustomUrlScheme;
+        }
+
+        if (!customUrlScheme.endsWith("://")) {
+            logger.warnf(
+                    "Custom URL scheme '%s' does not end with '://'. Defaulting to %s",
+                    customUrlScheme, defaultCustomUrlScheme);
+            return defaultCustomUrlScheme;
+        }
+
+        return customUrlScheme;
+    }
+
+    private static X509Certificate validateX5CCertificate(String certificate) {
+        if (StringUtils.isBlank(certificate)) {
+            return null;
+        }
+
+        try {
+            byte[] certBytes = Base64.getDecoder().decode(certificate);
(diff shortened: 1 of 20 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
```diff
@@ -148,47 +170,89 @@ private static String generateRandomString() {
...
         return new RequestObject()
                 .setIssuer(clientId)
-                .setResponseMode(ResponseMode.DIRECT_POST)
+                .setResponseMode(config.getResponseMode())
                 .setResponseUri(responseUri)
                 .setResponseType(ResponseType.VP_TOKEN)
                 .setClientId(clientId)
                 .setClientIdScheme(ClientIdScheme.X509_SAN_DNS)
                 .setNonce(nonce)
+                .setState(requestId)
```
Ogenbertrand, line 224: Always includes state, but OID4VPUserAuthEndpoint.java (line 152) only checks it when present. The new encrypted-response helper OID4VPBaseUserAuthEndpointTest.java (line 263) omits state entirely, and the HAIP happy-path test treats that as valid. That weakens request/response binding.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/VerifierConfig.java
```diff
@@ -0,0 +1,119 @@
...
+            return ResponseMode.fromValue(responseMode);
+        } catch (IllegalArgumentException e) {
+            String defaultResponseMode = SdJwtAuthenticatorFactory.RESPONSE_MODE_CONFIG_DEFAULT;
+            logger.warnf("Invalid response mode: %s. Defaulting to %s", responseMode, defaultResponseMode);
+            return ResponseMode.fromValue(defaultResponseMode);
+        }
+    }
+
+    private static String validateCustomUrlScheme(String customUrlScheme) {
+        String defaultCustomUrlScheme = SdJwtAuthenticatorFactory.RESPONSE_MODE_CONFIG_DEFAULT;
```
Ogenbertrand: This falls back to `RESPONSE_MODE_CONFIG_DEFAULT` instead of `CUSTOM_URL_SCHEME_CONFIG_DEFAULT`. A blank or malformed custom scheme therefore produces `direct_postauthorize?` instead of a valid `openid4vp://authorize?` link. Was it intentional?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/VerifierConfig.java
```diff
@@ -0,0 +1,119 @@
...
+
+        try {
+            byte[] certBytes = Base64.getDecoder().decode(certificate);
+            CertificateFactory cf = CertificateFactory.getInstance("X.509");
+            return (X509Certificate) cf.generateCertificate(new ByteArrayInputStream(certBytes));
+        } catch (CertificateException e) {
+            logger.warnf(e, "Invalid X5C certificate '%s'", certificate);
+            return null;
+        }
+    }
```
Ogenbertrand, line 99: This method isn't solid enough IMO,  It uses `Base64.getDecoder()` and only catches CertificateException, so wrapped PEM bodies or malformed base64 can throw `IllegalArgumentException` and surface as a 500. Other invalid cert inputs silently fall back to self-signed behavior which is risky.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointHAIPTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+        // Must use configured custom URL scheme
+        assertEquals(CUSTOM_URL_SCHEME, new URI(authRequest).getScheme());
+
+        // Request object must use configured response mode
+        assertEquals(ResponseMode.DIRECT_POST_JWT, requestObject.getResponseMode());
+
+        // Signed request object must embed access certificate in X5C header
+        ObjectNode authConfig = getAuthConfig();
+        String accessCertificate = authConfig.get(ACCESS_CERTIFICATE_CONFIG).asText();
```
Ogenbertrand, line 64: only checks that the configured certificate is copied into x5c, it never verifies that the signed JWT can actually be validated with that certificate. Because of that, the broken HAIP fixture test-realm-haip.json(see line 65) can still pass review even though it provisions an EC access cert while only configuring RSA signing material.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
```diff
@@ -85,30 +98,33 @@ public AuthorizationRequestService(KeycloakSession session) {
...
-                .setDcqlQuery(constrainer.generateDcqlQuery(queryMap))
-                // Kept for backward compatibility with Draft 20 wallets
-                .setPresentationDefinition(constrainer.generatePresentationDefinition(queryMap));
+        RequestObject requestObject = buildRequestObject(config, queryMap, requestId);
 
         // Sign request object
-        String requestObjectJwt = signRequestObject(requestObject);
+        X509Certificate certificate =
+                Optional.ofNullable(config.getAccessCertificate()).orElse(getSelfSignedCertificate());
+        String requestObjectJwt = signRequestObject(requestObject, certificate);
```
Ogenbertrand, line 123: Please reject accessCertificate unless its public key matches `signingKey.getPublicKey()` and its SAN matches the verifier client ID. Copying an unrelated cert into x5c makes the request object unverifiable.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/EphemeralKeyUtilsTest.java
```diff
@@ -0,0 +1,50 @@
...
+                Arguments.of(JWEConstants.ECDH_ES, JWEConstants.A128GCM),
+                Arguments.of(JWEConstants.ECDH_ES, JWEConstants.A192GCM),
+                Arguments.of(JWEConstants.ECDH_ES, JWEConstants.A256GCM),
+                Arguments.of(JWEConstants.ECDH_ES_A128KW, JWEConstants.A128GCM),
+                Arguments.of(JWEConstants.ECDH_ES_A192KW, JWEConstants.A256GCM));
+    }
+
+    @ParameterizedTest
+    @MethodSource("jweEncryptionAlgorithms")
+    public void testEncryptDecryptMessage(String jweAlg, String jweEncAlg) throws Exception {
```
Ogenbertrand, line 43: Add a round-trip assertion for toBase64String() and privateKeyFromBase64()

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPBaseUserAuthEndpointTest.java
```diff
@@ -236,16 +242,48 @@ protected HttpResponse sendAuthorizationResponseWithVPToken(
...
+     * @param sdJwtVpToken  the SD-JWT verifiable presentation token
+     * @param requestObject the request object containing the presentation definition
+     */
+    private List<BasicNameValuePair> prepareEncryptedOpenID4VPResponse(String sdJwtVpToken, RequestObject requestObject)
+            throws IOException {
+        // Build presentation submission
+        var vpTokenMap = prepareVpTokenMap(sdJwtVpToken, requestObject);
+        var respMap = Map.of(ResponseObject.VP_TOKEN_KEY, vpTokenMap);
+        String resp = JsonSerialization.writeValueAsString(respMap);
+        System.out.println(resp);
```
Ogenbertrand: The encrypted-response helper should include state, otherwise the test suite is validating non-compliant behavior. Also remove the stray System.out.println(resp) i don't think it's needed.

Ogenbertrand (changes requested):
Thanks for this PR @IngridPuppet, i left some comments for you could you check them!

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
IngridPuppet: Though it is encouraged to include a state. It is not mandatory to include IN the response payload. Because the response can now be received encrypted, request/response matching is now achieved by the state as a path parameter in the response endpoint.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/VerifierConfig.java
IngridPuppet: Definitely a mistake. Fixed, thanks.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/VerifierConfig.java
IngridPuppet: I agree. I updated to catch all Exceptions. I also switched from silent failure to throwing an IllegalStateException right away, otherwise we risk thinking all is working as expected.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointHAIPTest.java
IngridPuppet: This is all nice to have. For the current use case, we control everything, so nothing criticalll. It is fine postponing this check. I added a TODO to the codebase.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
IngridPuppet: Same as my previous response. Let's do it in a separate PR. It won't be straightforward to check because EC keys are generated.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPBaseUserAuthEndpointTest.java
IngridPuppet: Adding a state is not mandated. See https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-response-mode-direct_postjw

> Also remove the stray System.out.println(resp) i don't think it's needed.

It's a leftover. Removed.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/EphemeralKeyUtilsTest.java
IngridPuppet: Done

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
Ogenbertrand: I agree that we can send this fix in another PR, let's have it into an issue so we don'y forget.

Ogenbertrand (approved):
Please check this comment: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/52#discussion_r3094012342, asside we can merge this PR now, i ran the tests and they are all green as well.

IngridPuppet:
Thank you @Ogenbertrand for the review!

Ticket created: 
- https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/53

Please help triage it.

PR #56, opened 2026-04-21
Validate access certificate binding

Ogenbertrand:
closses: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/53

PR #57, opened 2026-04-21
Add support for Client ID scheme x509_hash

### LINKED ISSUE #54: Add support for client ID scheme x509_hash
Required by HAIP
https://bmi.usercontent.opencode.de/eudi-wallet/developer-guide/rp/PID_Presentation/#field-by-field-explanation

ETA: 1 day

IngridPuppet:
Closes #54


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
```diff
@@ -110,21 +100,27 @@ public AuthorizationContext createAuthorizationRequest(
         // the request object, so updated client metadata are picked up as intended.
         EphemeralKey encryptionKey = generateEncryptionKeyIfNeeded(config.getResponseMode());
 
+        // Discover signing key and client metadata
+        KeyWrapper signingKey = discoverSigningKey(config);
```
Ogenbertrand, line 103: This key is selected independently from accessCertificate, but under x509_hash the request must be signed by the private key matching the leaf certificate in x5c.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
```diff
@@ -110,21 +100,27 @@ public AuthorizationContext createAuthorizationRequest(
...
+        X509Certificate certificate = Optional.ofNullable(config.getAccessCertificate())
+                .orElseGet(() -> getSelfSignedCertificate(signingKey));
+        ClientMetadata clientMetadata =
+                discoveryService.getClientMetadata(config.getClientIdScheme(), certificate, encryptionKey);
```
Ogenbertrand: This fallback is non-compliant for HAIP x509_hash. when no configured certificate exists, this generates a self-signed certificate, but HAIP requires the request-signing certificate not to be self-signed. what were you reasons here ?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
```diff
@@ -170,6 +166,15 @@ private static String generateRandomString() {
         return Base64.getUrlEncoder().withoutPadding().encodeToString(randomBytes);
     }
 
+    /**
+     * Discovers signing key.
+     */
+    private KeyWrapper discoverSigningKey(VerifierConfig config) {
+        X509Certificate accessCertificate = config.getAccessCertificate();
+        boolean requireSelfSignedCert = accessCertificate == null;
```
Ogenbertrand, line 186: The boolean only toggles "certificate required" behavior, it does not ensure the chosen signer corresponds to accessCertificate when x509_hash is configured.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
Ogenbertrand, line 1: In the return `RequestObject` method of this class, if config selects x509_hash, client_id and the authorization URL use x509_hash, but the serialized request object still says x509_san_dns.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VerifierDiscoveryService.java
```diff
@@ -78,22 +96,38 @@ public KeyWrapper getSigningKey() {
         KeyWrapper key = keyManager.getActiveKey(realm, KeyUse.SIG, Algorithm.ES256);
 
         // Fall back to available key if ES256 is not available or its certificate missing.
-        if (key == null || key.getCertificate() == null) {
+        if (key == null || (requireSelfSignedCert && key.getCertificate() == null)) {
```
Ogenbertrand: This fallback is too loose for x509_hash. it picks any active signing key that fits broad criteria, but x509_hash requires the exact key pair corresponding to the configured verifier certificate.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VerifierDiscoveryService.java
```diff
@@ -78,22 +96,38 @@ public KeyWrapper getSigningKey() {
...
+        // Typically the hostname of the Keycloak server.
+        return session.getContext().getUri().getBaseUri().getHost();
+    }
+
+    private String getClientId(ClientIdScheme clientIdScheme, X509Certificate clientCertificate) {
+        // Compute client ID as per client ID scheme rules
+        String clientId =
+                switch (clientIdScheme) {
+                    case X509_SAN_DNS -> getDnsNameClientId();
+                    case X509_HASH -> X509HashUtils.computeX509Hash(clientCertificate);
```
Ogenbertrand, line 119: This hash computation is only correct if clientCertificate is the same leaf certificate whose private key signs the request object. otherwise the verifier identifier and request signature describe different entities.

Ogenbertrand (changes requested):
Thanks for this PR @IngridPuppet , could you check the few comments i left ?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/utils/X509HashUtils.java
```diff
@@ -0,0 +1,27 @@
+package io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.utils;
+
+import java.security.cert.X509Certificate;
+import java.util.Objects;
+import org.keycloak.common.util.Base64Url;
+import org.keycloak.crypto.JavaAlgorithm;
+import org.keycloak.jose.jws.crypto.HashUtils;
+
+public class X509HashUtils {
```
Awambeng, line 10: This class only contains `static` methods but can still be instantiated. Consider adding a private constructor to prevent instantiation.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/utils/X509HashUtils.java
```diff
@@ -0,0 +1,27 @@
...
+     * @see <a href="https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-defined-client-identifier-p">Defined Client Identifier Prefixes</a>
+     */
+    public static String computeX509Hash(X509Certificate cert) {
+        try {
+            byte[] encoded = Objects.requireNonNull(cert).getEncoded();
+            byte[] hashedOutput = HashUtils.hash(JavaAlgorithm.SHA256, encoded);
+            return Base64Url.encode(hashedOutput);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
```
Awambeng: This catch block is too broad and may swallow unrelated errors like NPE, encoding, or crypto issues, making debugging harder. Only `CertificateEncodingException` is expected here, so please narrow the catch to that specific exception

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VerifierDiscoveryService.java
```diff
@@ -78,22 +96,38 @@ public KeyWrapper getSigningKey() {
...
+                    default ->
+                        throw new IllegalArgumentException(
+                                "ClientIdScheme not supported: " + clientIdScheme.getValue());
```
Awambeng, line 122: Removing `default` makes this an exhaustive switch expression, so any future `ClientIdScheme` value added to the enum will produce a **compile-time error** rather than a silent runtime crash

Awambeng (changes requested):

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
Awambeng, line 1: On line 223 the clientId scheme is hardcoded. It should be read from configuration using `config.getClientIdScheme()`

Awambeng (changes requested):

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
IngridPuppet: Thank you for your review. The binding between the access certificate and the signing key is a concern of:
- https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/56

Any further improvement relating to this point should be handled separately for not supercharging this PR.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
IngridPuppet: We still want the plugin to work in a non-HAIP compliant mode. Not every use case will need an access certificate configured.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
IngridPuppet: The binding between the access certificate and the signing key is a concern of:
- https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/56

Any further improvement relating to this point should be handled separately for not supercharging this PR.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
IngridPuppet: Thank you. Updated to match config.

```java
.setClientIdScheme(config.getClientIdScheme())
```

IngridPuppet: Thank you. Updated to match config.

```java
.setClientIdScheme(config.getClientIdScheme())
```

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VerifierDiscoveryService.java
IngridPuppet: This was basically the behavior so far. The binding between the access certificate and the signing key is a concern of:
- https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/56

Any further improvement relating to this point should be handled separately for not supercharging this PR. In the meantime, the access certificate should be set in accordance with this selection rule.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VerifierDiscoveryService.java
IngridPuppet: For now, we have the binding between this hash client ID and the access certificate. The binding between the access certificate and the signing key is a concern of:
- https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/56

Any further improvement relating to this point should be handled separately for not supercharging this PR.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/utils/X509HashUtils.java
IngridPuppet: Thank you for your review. I added a private constructor to prevent instantiation. Thanks.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/utils/X509HashUtils.java
IngridPuppet: Narrowed to `CertificateEncodingException` only.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VerifierDiscoveryService.java
IngridPuppet: There are already values in enum ClientIdScheme that we do not want to support. For reference, options that can be configured are limited to these two schemes. For any future scheme under consideration, this exception will be caught by tests.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
IngridPuppet: Thank you. Updated to match config.

```java
.setClientIdScheme(config.getClientIdScheme())
```

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
Ogenbertrand: Oh ok !
It makes sense.

Awambeng (dismissed):

Ogenbertrand (dismissed):
After checking again, I confirm that the binding between the access certificate and the signing key is happening here already: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/56.
That said the changes of this PR LGTM, thanks @IngridPuppet.

Ogenbertrand (commented):
Great change, the changes lgtm, it now resolves the comments i left previously.
Just one question RSA certificates that was configured in the test ?

IngridPuppet:
> Just one question RSA certificates that was configured in the test ?

This path is already covered by HAIP tests. When an access certificate is configured, we should be able to use an ES256 Key without self-signed certificate.

Ogenbertrand:
> > Just one question RSA certificates that was configured in the test ?
> 
> This path is already covered by HAIP tests. When an access certificate is configured, we should be able to use an ES256 Key without self-signed certificate.

Oh yeh !
Thanks for clarifying.

Ogenbertrand (approved):
PR LGTM

PR #58, opened 2026-04-22
chore: add OID4VCI request patch and suppress metadata encryption parameters

Awambeng:
This PR introduces the following changes:

- Added `PatchedOID4VCIssuerEndpoint` to map `credential_configuration_id` → `credential_identifier`
- Removed `credential_response_encryption` and `credential_request_encryption` from issuer metadata
- Added unit tests for the endpoint mapping logic
- Updated metadata provider tests to reflect removed encryption fields

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/patch/issuance/PatchedOID4VCLoginProtocolFactory.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/patch/issuance/PatchedOID4VCLoginProtocolFactory.java
@@ -1,0 +1,22 @@
+package io.github.adorsysgis.keycloak.protocol.oid4vc.patch.issuance;
+
+import org.keycloak.events.EventBuilder;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.protocol.oid4vc.OID4VCLoginProtocolFactory;
+
+/**
+ * Overrides the default {@link OID4VCLoginProtocolFactory} with a higher priority
+ * to provide the {@link PatchedOID4VCIssuerEndpoint}.
+ */
+public class PatchedOID4VCLoginProtocolFactory extends OID4VCLoginProtocolFactory {
+
+    @Override
+    public Object createProtocolEndpoint(KeycloakSession session, EventBuilder event) {
+        return new PatchedOID4VCIssuerEndpoint(session);
+    }
+
+    @Override
+    public int order() {
+        return super.order() + 10; // Higher than default -> this factory wins
+    }
+}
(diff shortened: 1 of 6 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/patch/issuance/PatchedOID4VCLoginProtocolFactory.java
```diff
@@ -0,0 +1,22 @@
...
+
+    @Override
+    public Object createProtocolEndpoint(KeycloakSession session, EventBuilder event) {
+        return new PatchedOID4VCIssuerEndpoint(session);
+    }
+
+    @Override
+    public int order() {
+        return super.order() + 10; // Higher than default -> this factory wins
+    }
```
IngridPuppet, line 21: I assume Alex's code uses 10. Let's use 9 here, so his code always take precedence.

IngridPuppet (commented):
Just one minor comment

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/patch/issuance/PatchedOID4VCLoginProtocolFactory.java
Awambeng: Done

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/patch/issuance/PatchedOID4VCIssuerEndpointTest.java
```diff
@@ -0,0 +1,36 @@
...
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+import org.junit.jupiter.api.Test;
+import org.keycloak.protocol.oid4vc.model.CredentialRequest;
+import org.keycloak.util.JsonSerialization;
+
+public class PatchedOID4VCIssuerEndpointTest {
+
+    @Test
+    public void testPatchWalletRequest() throws Exception {
```
forkimenjeckayang, line 13: Nice coverage for the mapping path. Could we add 2 defensive cases to lock current failure-safe behavior?

- invalid JSON input should be forwarded unchanged (no throw)
- whitespace-only payload (e.g. " ") should be returned unchanged.

This matches current implementation intent (catch -> forward as-is) and helps prevent regressions.

forkimenjeckayang (commented):
Minor concern but not a blocker for the PR to be merged . WDYT ?

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/patch/issuance/PatchedOID4VCIssuerEndpointTest.java
Awambeng: Done.

IngridPuppet (approved):
Thx

forkimenjeckayang (approved):

PR #59, opened 2026-04-22
Clean up metadata exposed in request object

IngridPuppet:
Regarding metadata exposed in request object, this PR:
- Explicitly advertises supported encryption algorithms (`encrypted_response_enc_values_supported`). Confirmed by EphemeralKeyUtilsTest.
- Only advertises asymetric algorithms in vpFormat.
- Add configuration for exposing `presentation_definition` XOR newer `dcql_query`. Not both.
- Removes deprecated clientId field in clientMetadata.

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/VerifierConfig.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/VerifierConfig.java
@@ -4,6 +4,7 @@ import com.apicatalog.jsonld.StringUtils;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.authenticator.SdJwtAuthRequirements;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.authenticator.SdJwtAuthenticatorFactory;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model.ClientIdScheme;
+import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model.QueryLanguage;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model.ResponseMode;
 import java.io.ByteArrayInputStream;
 import java.security.cert.CertificateFactory;
@@ -26,6 +27,7 @@ public class VerifierConfig {
     private final SdJwtAuthRequirements authRequirements;
 
     private final ClientIdScheme clientIdScheme;
+    private final QueryLanguage queryLanguage;
     private final ResponseMode responseMode;
     private final String authReqUrlScheme;
     private final X509Certificate accessCertificate;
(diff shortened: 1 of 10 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/VerifierConfig.java
Awambeng, line 1: Maybe it would be good to make this class immutable to avoid unexpected side effects. WDYT?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VerifierDiscoveryService.java
```diff
@@ -138,9 +136,6 @@ private SdGenericFormat getSdJwtVpFormat() {
     }
 
     private List<String> getSupportedSignatureAlgorithms() {
-        return session.getKeycloakSessionFactory()
-                .getProviderFactoriesStream(SignatureProvider.class)
-                .map(ProviderFactory::getId)
-                .toList();
+        return CryptoUtils.getSupportedAsymmetricSignatureAlgorithms(session);
```
Awambeng, line 139: I’m worried this list might grow and make the QR code harder to scan. Would it make sense to use an allowlist instead to keep it smaller?

Awambeng (changes requested):
@IngridPuppet thanks for the PR! I’ve left a few comments.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/VerifierConfig.java
IngridPuppet: You're right that immutability is wanted. Notice that all fields are final. Consequently, no setters are available. So it's technically already immutable. If your suggestion is to turn the class into a record, I'm afraid that will hurt readability and cause too many renamings across the database at this point.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VerifierDiscoveryService.java
IngridPuppet: The list of supported asymmetric signature algorithms is very stable and unlikely to grow to the point of making QR codes difficult to scan. Also notice that this update actually reduces this list of algorithms. It's an improvement.

```java
[PS384, RS384, EdDSA, ES384, ES256, RS256, ES512, PS256, PS512, RS512]
```

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/VerifierConfig.java
Awambeng: Thanks for the clarification
I didn’t look closely enough and just assumed it was mutable because of the getters

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VerifierDiscoveryService.java
Awambeng: Looks great, thank you!

Awambeng (approved):
LGTM

PR #60, opened 2026-04-23
Further clean up metadata exposed in request object

IngridPuppet:
- Add `credential_sets` to DCQL
- Add `ECDH_ES` to encryption keys
- Rename `vp_formats` to `vp_formats_supported`
- Restore ability to expose both DCQL and DIF presentation exchange requests
- Reduce the list of advertised signing and encryption algorithms (could be causing incompatibility issues)

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VerifierDiscoveryService.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VerifierDiscoveryService.java
@@ -33,7 +33,7 @@ public class VerifierDiscoveryService {
 
     private static final Logger logger = Logger.getLogger(VerifierDiscoveryService.class);
 
-    public static final List<String> SUPPORTED_ENC_ALGS = List.of(JWEConstants.A128GCM);
+    public static final List<String> SUPPORTED_ENC_ALGS = List.of(JWEConstants.A256GCM);
 
     private final KeycloakSession session;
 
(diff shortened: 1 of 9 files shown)


Ogenbertrand (dismissed):
I understand that this changes is for compactibility for now, are we going to revert this changes once our wallet start supporting DCQL ?
That said the PR LGTM.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VerifierDiscoveryService.java
```diff
@@ -136,6 +134,7 @@ private SdGenericFormat getSdJwtVpFormat() {
     }
 
     private List<String> getSupportedSignatureAlgorithms() {
-        return CryptoUtils.getSupportedAsymmetricSignatureAlgorithms(session);
+        // TODO: CryptoUtils.getSupportedAsymmetricSignatureAlgorithms(session);
```
Ogenbertrand, line 137: Don't you also think this TODO is necessary now ?

IngridPuppet:
> I understand that this changes is for compactibility for now, are we going to revert this changes once our wallet start supporting DCQL ?

Which one in particular? The option to select the query language? That can stay for quite some time in my opinion.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VerifierDiscoveryService.java
IngridPuppet: We fear this is causing issues. We need to check first with ES256.

Ogenbertrand:
> > I understand that this changes is for compactibility for now, are we going to revert this changes once our wallet start supporting DCQL ?
> 
> Which one in particular? The option to select the query language? That can stay for quite some time in my opinion.

It's fine then, my concern was just that it's against the spec.

Ogenbertrand (approved):
PR LGTM

IngridPuppet:
Thank you for the review! @Ogenbertrand

PR #70, opened 2026-05-05
Add support for same-device flow in browser-facing authentication

### LINKED ISSUE #49: Add support for externally signed certificates

IngridPuppet:
Closes https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/49 (third and last part)

https://github.com/user-attachments/assets/15753412-c29e-4a67-8b5d-42a5c3fc8ef6

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
@@ -30,6 +30,7 @@ import jakarta.ws.rs.core.Response;
 import java.io.IOException;
 import java.net.URI;
 import java.security.interfaces.ECPrivateKey;
+import java.util.Objects;
 import org.jboss.logging.Logger;
 import org.keycloak.OAuth2Constants;
 import org.keycloak.OAuthErrorException;
@@ -218,12 +219,13 @@ public class OID4VPUserAuthEndpoint extends OID4VPUserAuthEndpointBase implement
         logger.debug("Handling redirect callback for same-device authentication...");
 
         AuthenticationSessionModel authSession = null;
+        AuthenticationSessionStore authStore;
         AuthorizationContext authContext;
 
         try {
             authSession = this.recoverAuthenticationSession(responseCode);
-            authContext =
-                    new AuthenticationSessionStore(authSession).getAuthorizationContextByResponseCode(responseCode);
+            authStore = new AuthenticationSessionStore(authSession);
+            authContext = authStore.getAuthorizationContextByResponseCode(responseCode);
         } catch (IllegalArgumentException e) {
             String msg = "Authorization context not found for response code: " + responseCode;
             logger.error(msg, e);
@@ -231,11 +233,18 @@ public class OID4VPUserAuthEndpoint extends OID4VPUserAuthEndpointBase implement
         }
 
         // Check cookie-tracked session is consistent with this redirection attempt
-        if (!matchesCookieTrackedAuthSession(authContext.getParentAuthSessionId(), authContext.getLoginActionUrl())) {
+        if (!matchesCookieTrackedAuthSession(
+                authContext.getParentAuthSessionId(), Objects.requireNonNull(authContext.getLoginActionUrl()))) {
             String msg = "Authentication session does not match cookie-tracked session";
             logger.error(msg);
             return ErrorPage.error(session, authSession, Response.Status.BAD_REQUEST, msg);
         }
+
+        // Invalidate current response code to prevent reuse.
+        // The field is not voided for it is used as a marker of same-device flows.
+        String newRandomResponseCode = AuthorizationRequestService.generateSessionBoundId(authSession);
+        authContext.setResponseCode(newRandomResponseCode);
+        authStore.storeAuthorizationContext(authContext);
 
         // Build redirect URI
         URI redirectUri = KeycloakUriBuilder.fromUri(authContext.getLoginActionUrl())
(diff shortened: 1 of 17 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
```diff
@@ -187,8 +193,57 @@ public Response processAuthorizationResponse(
...
+            return ErrorPage.error(session, authSession, Response.Status.BAD_REQUEST, msg);
+        }
+
+        // Build redirect URI
+        URI redirectUri = KeycloakUriBuilder.fromUri(authContext.getLoginActionUrl())
+                .queryParam(OAuth2Constants.CODE, authContext.getAuthorizationCode())
+                .build();
+
+        // Return redirect response
+        return Response.status(Response.Status.FOUND).location(redirectUri).build();
```
forkimenjeckayang, line 255: `response_code` appears to be reusable: we validate match but never consume/invalidate it after successful callback. Given `direct_post` reference designs (draft-20/final-1.0) model `response_code` as one-time bridge secret, this should be invalidated after first successful use to prevent replay.
Could we clear or mark `responseCode` as consumed and persist context before issuing redirect?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthenticationSessionStore.java
```diff
@@ -66,6 +66,19 @@ public AuthorizationContext getAuthorizationContextByTransactionId(String transa
...
+     * Retrieves authorization context by response code from the authentication session.
+     */
+    public AuthorizationContext getAuthorizationContextByResponseCode(String responseCode) {
+        AuthorizationContext authContext = getAuthorizationContext();
+        if (!Objects.equals(authContext.getResponseCode(), responseCode)) {
+            throw new IllegalArgumentException(
+                    "Authorization context does not match the provided response code: " + responseCode);
+        }
+
+        return authContext;
```
forkimenjeckayang, line 79: This accessor only checks equality and returns context; it does not enforce one-time semantics.
For same-device callback hardening, can we enforce consume-on-read (or explicit consume step) so reused `response_code` fails?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointBase.java
```diff
@@ -141,6 +165,40 @@ protected AuthenticationSessionModel createAuthSession(ClientModel client) {
         return authSession;
     }
 
+    /**
+     * Checks if session parameters match current cookie-tracked authentication session.
+     */
+    protected boolean matchesCookieTrackedAuthSession(String authSessionId, String loginActionUrl) {
+        String query = URI.create(loginActionUrl).getQuery();
+        List<NameValuePair> params = URLEncodedUtils.parse(query, StandardCharsets.UTF_8);
```
forkimenjeckayang, line 174: `URI.create(loginActionUrl)` can throw for null/invalid values and may cause a 500 path.
Can we harden this by validating input + try/catch, then return `false` (session mismatch) so callback responds with controlled 400 instead of internal error?

forkimenjeckayang (changes requested):
Thank you for the work on this @IngridPuppet 
A few comments to check

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
IngridPuppet: Thank you for your review. I've updated the flow so response code is single-use. It is indeed a good security practice. However, could you share the excerpt from the spec where this is mandated? It could sometimes cause practical difficulties. Let's see how it goes.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthenticationSessionStore.java
IngridPuppet: I guess this is addressed by the update in response to the previous point.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointBase.java
IngridPuppet: With this strict philosophy, you'll quickly end up with bloated code validating input at every method start and with try/catch every couple of lines. Here, loginActionUrl is not user-provided; we know it's logically not null, given the conditions that led to calling this method. To help static analysis, I've annotated the method argument with JSpecify @NonNull. I think we should recommend these null-checking annotations going forward over null checks just to throw another exception.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
forkimenjeckayang: Thanks for the update 
Spec basis:

OID4VP Final 1.0 [§14.2 (Session Fixation)](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-session-fixation) requires the Response URI to return a fresh secret (Response Code) in `redirect_uri` and to require that Response Code when the frontend fetches the Authorization Response.

[§14.3.3 (Protection of Authorization Response Data)](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-protection-of-the-authoriza) requires protections on the internal fetch interface, and explicitly lists the two-random-values approach (state + retrieval secret).

[§13.3 reference design ](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-response-mode-direct_post-2) shows `response_code` as the bridge secret bound to one response and consumed in the completion flow.

So while “single-use” is most explicit in the reference design/security rationale, the MUST-level requirements in the spec support this implementation pattern directly.

forkimenjeckayang (approved):
LGTM

IngridPuppet:
@forkimenjeckayang - Thank you for the review.

PR #78, opened 2026-05-06
Implement Final 1.0 direct_post.jwt handling and conformance checks

### LINKED ISSUE #69: Decide and Implement direct_post.jwt Strategy
Description: Either explicitly declare unsupported or implement final 1.0-compliant encrypted JWT response handling for direct_post.jwt.

### LINKED ISSUE #74: Encryption JWKs (client_metadata.jwks)
Description: Publish JWKs for response encryption with alg and unique kid per key. These keys must not be used to verify authorization request signatures (Final restriction).
Spec:
https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-encrypted-responses and client metadata rules
https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-new-parameters

forkimenjeckayang:
- Add strict `direct_post.jwt` encrypted response validation in the OID4VP endpoint, including JWE envelope checks (`alg`, `enc`, `kid`) before decryption.
- Bind response encryption key selection to request context (`expectedEncryptionKid`) and enforce header/JWK consistency.
- Align verifier metadata with runtime behavior by advertising and consuming supported encrypted response algorithms (`encrypted_response_alg_values_supported`, `encrypted_response_enc_values_supported`).
- Extend test utilities and HAIP endpoint tests to cover valid flow and rejection paths (unsupported `alg`, unsupported `enc`, invalid `kid`, malformed JWE).

## Spec Alignment (Final 1.0)
- `8.3 Encrypted Responses`
- `8.3.1 Response Mode direct_post.jwt`
- `5.1 client_metadata` requirements relevant to encrypted response handling
- `14.3 direct_post` response handling security considerations

closes https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/69 https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/74


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ClientMetadata.java
```diff
@@ -25,6 +25,9 @@ public class ClientMetadata {
     @JsonProperty("encrypted_response_enc_values_supported")
     private List<String> encryptedResponseEncValuesSupported;
 
+    @JsonProperty("encrypted_response_alg_values_supported")
+    private List<String> encryptedResponseAlgValuesSupported;
+
```
Ogenbertrand: remove the non-standard metadata field from serialized client_metadata, `encrypted_response_alg_values_supported` is not a final 1.0 client metadata parameter

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
Ogenbertrand, line 1: Please verify the `processAuthorizationResponse` method in this class, check line 156, the he endpoint requires response to be present whenever an ephemeral key exists. accept and process unencrypted error, `error_description`, and state when response is absent, even for `direct_post.jwt`. Add a regression test for that path.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
```diff
@@ -384,6 +391,93 @@ private ResponseObject decryptResponse(String encryptedResponse, String ephemera
...
+        }
+
+        JWK[] keys = resolveEncryptionKeys(clientMetadata);
+        if (keys == null || keys.length == 0) {
+            throw new IllegalArgumentException("jwe_key_unavailable");
+        }
+        String expectedKid = authorizationContext.getExpectedEncryptionKid();
+        if (kid == null) {
+            throw new IllegalArgumentException("jwe_kid_missing");
+        }
```
Ogenbertrand: I feel this is too stricker than the spec, it says “If the selected public key contains a kid parameter, the JWE MUST include the same value...” the block above enforces it.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
Ogenbertrand, line 1: Can we also avoid referencing spec versions, as they might change again.

Ogenbertrand (changes requested):

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
forkimenjeckayang: Thanks for the note.  In `client_metadata.jwks` each JWK MUST have a unique kid, and for encrypted responses the JWE header must carry the selected key’s kid. Since this endpoint relies on `client_metadata.jwks` and key binding to `expectedEncryptionKid`, requiring kid here is intentional and aligns with spec-conformant direct_post.jwt handling.
See: https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-new-parameters

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
forkimenjeckayang: Thanks for the suggestion, removed for now, but we can keep OpenID4VP Final 1.0 references since this flow is intentionally Final-only (not draft), and the reference helps make compliance intent explicit.

Ogenbertrand (dismissed):
Thanks for considering my comments @forkimenjeckayang !
I confirm that the logic meetups with the requirements of the OID4VP spec.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/dto/AuthorizationContext.java
```diff
@@ -94,6 +94,12 @@ public class AuthorizationContext {
     @JsonProperty("ephemeral_key")
     private String ephemeralKey;
 
+    /**
+     * Expected key identifier of the ephemeral response-encryption key.
+     */
+    @JsonProperty("expected_encryption_kid")
+    private String expectedEncryptionKid;
```
IngridPuppet: Could you introduce a new model that will encapsulate both? My point is to avoid exploding fields in this class.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
```diff
@@ -439,6 +471,93 @@ private ResponseObject decryptResponse(String encryptedResponse, String ephemera
...
+        // The OID4VP spec defines encrypted_response_enc_values_supported in client_metadata, but not
+        // encrypted_response_alg_values_supported. We enforce our supported key management algs server-side.
+        return List.of(JWEConstants.ECDH_ES);
+    }
+
+    private static List<String> resolveAllowedEnc(ClientMetadata clientMetadata) {
+        List<String> allowedEnc =
+                clientMetadata != null ? clientMetadata.getEncryptedResponseEncValuesSupported() : null;
+        // Use the spec default when verifier did not explicitly advertise supported enc values.
+        return (allowedEnc == null || allowedEnc.isEmpty()) ? List.of(JWEConstants.A128GCM) : allowedEnc;
```
IngridPuppet: This code is very repetitive. Also, consider using Optional.ofNullable.

```suggestion
        return
                clientMetadata != null ? clientMetadata.getEncryptedResponseEncValuesSupported() : SUPPORTED_ENC_ALGS;
        // Use the spec default when verifier did not explicitly advertise supported enc values.
```

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
```diff
@@ -439,6 +471,93 @@ private ResponseObject decryptResponse(String encryptedResponse, String ephemera
...
+
+        byte[] decodedHeader = Base64.getUrlDecoder().decode(parts[0]);
+        JsonNode jweHeader = JsonSerialization.mapper.readTree(decodedHeader);
+        String alg = getTextClaim(jweHeader, "alg");
+        String enc = getTextClaim(jweHeader, "enc");
+        String kid = getTextClaim(jweHeader, "kid");
+
+        if (alg == null || enc == null) {
+            throw new IllegalArgumentException("jwe_header_invalid: missing alg or enc");
+        }
```
IngridPuppet: There is a JWE stack in Keycloak. Please consider reusing it for cleaner code.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
```diff
@@ -439,6 +471,93 @@ private ResponseObject decryptResponse(String encryptedResponse, String ephemera
...
+    }
+
+    private static JWK resolveSelectedKey(JWK[] keys, String expectedKid) {
+        for (JWK key : keys) {
+            if (expectedKid.equals(key.getKeyId())) {
+                return key;
+            }
+        }
+        throw new IllegalArgumentException("jwe_key_unavailable");
+    }
```
IngridPuppet: This is quite redundant. The actual source of truth should be the key in the request object. Any limitation encountered?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
```diff
@@ -439,6 +471,93 @@ private ResponseObject decryptResponse(String encryptedResponse, String ephemera
...
+
+    private static String getTextClaim(JsonNode node, String claimName) {
+        JsonNode claim = node.get(claimName);
+        return claim != null && claim.isTextual() ? claim.asText() : null;
+    }
```
IngridPuppet: We always want to keep this Endpoint class quite short. All these methods are related to a specific logic that can be encapsulated into another class and just reused here. Because they span over quite a lot of lines, It is definitely preferable to relocate them. This also makes testing easier, allowing not be extensive in upper layers.

Also, double check the error messages you throw with exceptions. They should be optimised for human readability. For example "jwe_key_unavaible" -> "Key id %s matches no available keys for JWE encryption".

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointHAIPTest.java
```diff
@@ -147,6 +156,90 @@ public void shouldRejectUnencryptedResponses() throws Exception {
...
+        HttpResponse response = httpClient.execute(httpPost);
+        assertEquals(HttpStatus.SC_OK, response.getStatusLine().getStatusCode());
+
+        HttpResponse statusResponse = fetchAuthenticationStatus(authContext.getTransactionId());
+        assertEquals(HttpStatus.SC_OK, statusResponse.getStatusLine().getStatusCode());
+        AuthorizationContext status = parseAuthorizationContext(statusResponse);
+        assertEquals(AuthorizationContextStatus.ERROR, status.getStatus());
+        assertEquals(ProcessingError.VP_TOKEN_AUTH_ERROR, status.getError());
+        assertTrue(status.getErrorDescription().contains("Wallet returned error: access_denied"));
+        assertTrue(status.getErrorDescription().contains("wallet canceled presentation"));
```
IngridPuppet: Can't this be just one assertion?

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointHAIPTest.java
```diff
@@ -147,6 +156,90 @@ public void shouldRejectUnencryptedResponses() throws Exception {
...
+                new BasicNameValuePair(ResponseObject.STATE_KEY, authContext.getRequestId()))));
+
+        HttpResponse response = httpClient.execute(httpPost);
+        assertEquals(HttpStatus.SC_OK, response.getStatusLine().getStatusCode());
+
+        HttpResponse statusResponse = fetchAuthenticationStatus(authContext.getTransactionId());
+        assertEquals(HttpStatus.SC_OK, statusResponse.getStatusLine().getStatusCode());
+        AuthorizationContext status = parseAuthorizationContext(statusResponse);
+        assertEquals(AuthorizationContextStatus.ERROR, status.getStatus());
+        assertEquals(ProcessingError.VP_TOKEN_AUTH_ERROR, status.getError());
```
IngridPuppet: No VP token was received by Keycloak. Is this the appropriate error type?

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointHAIPTest.java
```diff
@@ -147,6 +156,90 @@ public void shouldRejectUnencryptedResponses() throws Exception {
...
+        // Build minimal encrypted response payload
+        var dcql = requestObject.getDcqlQuery();
+        String credentialId = dcql.getCredentials().getFirst().getId();
+        var vpTokenMap = Map.of(credentialId, List.of("sd-jwt-vp-token"));
+        String payload = JsonSerialization.writeValueAsString(Map.of(ResponseObject.VP_TOKEN_KEY, vpTokenMap));
+
+        // Encrypt with requested variations
+        JWK encJwk = requestObject.getClientMetadata().getJwks().getKeys()[0];
+        var encKey =
+                (java.security.interfaces.ECPublicKey) JWKParser.create(encJwk).toPublicKey();
```
IngridPuppet: Import the class

IngridPuppet (changes requested):
Hello @forkimenjeckayang - I went through the PR and left some comments. Please could you check?

forkimenjeckayang:
Thank you for detailed review @IngridPuppet 
Please do check the updated changes when you get the chance

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/dto/AuthorizationContext.java
```diff
@@ -218,23 +214,46 @@ public AuthorizationContext setRequestObjectJwt(String requestObjectJwt) {
...
+    private boolean emptyEncryptionKid() {
+        return responseEncryption == null || responseEncryption.getExpectedEncryptionKid() == null;
+    }
+
+    private ResponseEncryptionContext ensureResponseEncryption() {
+        if (responseEncryption == null) {
+            responseEncryption = new ResponseEncryptionContext();
+        }
+        return responseEncryption;
+    }
```
IngridPuppet: It is not sustainable to maintain this logic. It is fine to access all under sub object ResponseEncryptionContext.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/dto/AuthorizationContext.java
```diff
@@ -89,16 +90,11 @@ public class AuthorizationContext {
     private String requestObjectJwt;
 
     /**
-     * An optional ephemeral key for encrypting responses.
+     * Ephemeral response-encryption key material for {@code direct_post.jwt}. Unwrapped in JSON so
+     * {@code ephemeral_key} and {@code expected_encryption_kid} remain at the same level as other context fields.
```
IngridPuppet: The point is precisely not to have this unwrapping so fields do not explode unnested.

IngridPuppet (commented):
Just one minor comment

forkimenjeckayang:
@IngridPuppet . I've  reverted chnages on AuthorizationContext , removed unwrapping/delegation,  kid binding now comes from client_metadata.jwks in the signed request. Hope this addresses final concerns

IngridPuppet (approved):
Thank you for the update. I appreciate the new concise version, with a single source of truth for kid.

PR #79, opened 2026-05-07
Align client_id prefix handling with strict KB-JWT audience equality

### LINKED ISSUE #65: Align Client Identifier Handling with Prefix Model
Description: Use full prefixed client_id in requests and enforce strict audience equality for key-binding JWT checks.
Specification reference:
https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-client-identifier-prefix-an
https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-always-use-the-full-client-
Recommended approach: Centralize client-id construction (single service) and reuse for both request object + KB-JWT audience validator.

forkimenjeckayang:
- Enforce strict equality for key-binding JWT `aud` validation against the expected verifier `client_id`.
- Remove permissive audience matching that allowed additional scheme/prefix variants.
- Keep client identifier handling aligned with full prefixed `client_id` semantics.

## Spec Alignment
- OpenID4VP Final 1.0, Client Identifier Prefix model (Section 5.9).
- OpenID4VP Final 1.0, Always use full Client Identifier (Section 14.8).
- Key-binding audience binding requirements (`aud` bound to verifier identifier, non-DC API flow).

closes https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/65

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthRequirements.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthRequirements.java
@@ -131,12 +131,8 @@ public class SdJwtAuthRequirements {
     }
 
     private static ClaimCheck buildAudClaimCheck(String expectedKbJwtAud) {
-        // Some wallets prepend a scheme to the expected audience. We accept any such scheme.
-        String regex = String.format("([^:]+:)?%s", Pattern.quote(expectedKbJwtAud));
-        Pattern expectedPattern = Pattern.compile(regex);
-        return new ClaimCheck(JsonWebToken.AUD, expectedKbJwtAud, (expectedAud, aud) -> expectedPattern
-                .matcher(aud)
-                .matches());
+        // Final 1.0 requires using the full Client Identifier, including prefix, in proof bindings.
+        return new ClaimCheck(JsonWebToken.AUD, expectedKbJwtAud, String::equals);
     }
 
     private List<String> parseMultiStr(String str) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthRequirements.java
Ogenbertrand, line 1: Can you add a test here, i'll continue reviewing once the logic is tested, i have a feeling the changes doesn't fully addresses the scope of the task, we should sync if possible after the test.

Ogenbertrand (commented):
Thanks for the PR @forkimenjeckayang 
Could you check my comment below.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthRequirements.java
forkimenjeckayang: Added test coverage for strict aud equality (including unprefixed host as invalid) in `shouldFailAuthentication_InvalidKbJwt_InvalidAud`; targeted test passes.

Awambeng (approved):
@forkimenjeckayang Thanks for the PR, it looks good overall.

During the Sprind test, I remember we implemented support for client identifiers in the wallet, and the logic was stripping the prefix from the client identifier. According to the OpenID4VP spec, the full client_id (including prefix) must be preserved and used consistently, so this behavior would not be compliant.

It might be worth creating a follow-up ticket on the wallet side to ensure we always handle the full client identifier without modification.

Ogenbertrand (approved):
LGTM

forkimenjeckayang:
> @forkimenjeckayang Thanks for the PR, it looks good overall.
> 
> During the Sprind test, I remember we implemented support for client identifiers in the wallet, and the logic was stripping the prefix from the client identifier. According to the OpenID4VP spec, the full client_id (including prefix) must be preserved and used consistently, so this behavior would not be compliant.
> 
> It might be worth creating a follow-up ticket on the wallet side to ensure we always handle the full client identifier without modification.

definitely there will be tickets to update the wallet for full oid4vp 1.0 support

PR #80, opened 2026-05-07
fix(oid4vp): enforce response_uri prefix rules and wallet JSON response contract

### LINKED ISSUE #77: response_uri and HTTP response to the wallet
Description: Validate response_uri under the same kind of rules as redirect_uri for the active client-identifier prefix. When the plugin responds on the response-uri channel, return HTTP 200, Content-Type: application/json, and a JSON body as required.
Spec:
https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-response-mode-direct_post
https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-validation-of-the-response-
https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-protection-of-the-response-
https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-response-mode-direct_post-2

forkimenjeckayang:
- validate `response_uri` with redirect-uri-like rules for the active client identifier scheme
- enforce prefix-aware checks for supported schemes (`x509_san_dns`, `x509_hash`) and reject unsupported schemes early
- ensure wallet callback endpoint behavior matches Final 1.0 expectations (HTTP 200 + JSON response contract on successful processing)
- add/extend tests to verify response-uri host consistency and JSON response semantics

## Spec alignment
- OpenID4VP Final 1.0: `direct_post` response mode requirements
- `response_uri` parameter validation under Section 5.9 prefix model semantics
- Section 14.3.1 (Validation of the Response URI)
- Verifier endpoint reply requirements for wallet callback channel

closes https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/77


Ogenbertrand (dismissed):
The changes looks ok to me.
Thank you @forkimenjeckayang

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
```diff
@@ -271,6 +274,45 @@ private RequestObject buildRequestObject(
...
+            default ->
+                throw new IllegalArgumentException(
+                        "Unsupported client_id scheme for response_uri validation: " + clientIdScheme.getValue());
+        }
+    }
+
+    private static boolean isLoopbackHost(String host) {
+        return "localhost".equalsIgnoreCase(host) || "127.0.0.1".equals(host) || "::1".equals(host);
+    }
+
```
IngridPuppet: We create the response URI ourselves. There is no need to run this validation after we're done creating it. If you believe any of these rules could be violated, an exception must be thrown before creating the URI, and for a reason that is not under our control.

If there are doubts something might incorrectly implemented, that's what tests are for. Additionally, it is usually fine to bypass the http and loopback host rules because Keycloak will naturally makes this not possible in prod mode. 

Do you have a scenario in mind in which any of these validated rules could be broken?

IngridPuppet (changes requested):
I'm concerned about the rationale behind these changes. Please could you check my comment below?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
forkimenjeckayang: Applied @IngridPuppet 
Removed `validateResponseUri()` since we construct `response_uri` ourselves. Kept the integration tests for wallet callback `application/json` and `response_uri` host alignment with `x509_san_dns client_id`.

IngridPuppet (approved):
LGTM

PR #84, opened 2026-05-12
Implement wallet error response handling at response_uri

### LINKED ISSUE #68: Implement Wallet Error Response Handling at response_uri
Description: Accept and persist OAuth/OpenID4VP error responses posted by wallet (instead of treating them as malformed VP submissions).
Specification reference: [Final 1.0 section 8.5](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-error-response)
Recommended approach: Add alternate form-param branch (error, error_description, state) and map into existing auth-context error status model.

forkimenjeckayang:
- Add explicit wallet error-response branch in `OID4VP` response processing to accept OAuth/OpenID4VP form params (`error`, optional `error_description`, `state`) instead of treating them as malformed VP submissions.
- Persist wallet-posted errors into the existing authorization context error model (`status=error`, mapped `ProcessingError`, `error_description`) so polling clients can retrieve final failure state.
- Enforce request-shape validation for error posts by rejecting mixed payloads that include both error params and VP response params.
## Spec alignment
- Aligns with OpenID4VP Final 1.0 error response handling (Section 8.5 context) and `direct_post` response endpoint behavior.
- Supports unencrypted wallet error posts even when encrypted VP responses are configured, consistent with spec guidance for error fallback.

closes https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/68


Awambeng (dismissed):
@forkimenjeckayang Thank you for the PR! The changes look good to me. However, we might want to add support for the optional `error_uri` parameter if the plugin needs to expose external documentation links provided by the wallet. This is optional and up to you if you think it’s necessary.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointTest.java
```diff
@@ -373,6 +373,47 @@ public void shouldFailAuthentication_InvalidSdJwtVPToken_Empty() throws Exceptio
         assertTrue(errorRep.getErrorDescription().contains("Could not parse SD-JWT VP token contained in `vp_token`"));
     }
 
+    @Test
+    public void shouldPersistWalletErrorResponse() throws Exception {
+        AuthorizationContext authContext = requestAuthorizationRequest();
+        RequestObject requestObject = resolveRequestObject(authContext.getAuthorizationRequest());
```
IngridPuppet: Does this test anything new compared to `shouldAcceptUnencryptedWalletErrorResponse`?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/dto/ProcessingError.java
```diff
@@ -10,7 +12,20 @@
...
+    ACCESS_DENIED("access_denied"),
+    UNAUTHORIZED_CLIENT("unauthorized_client"),
+    UNSUPPORTED_RESPONSE_TYPE("unsupported_response_type"),
+    SERVER_ERROR("server_error"),
+    TEMPORARILY_UNAVAILABLE("temporarily_unavailable"),
+    VP_FORMATS_NOT_SUPPORTED("vp_formats_not_supported"),
+    INVALID_REQUEST_URI_METHOD("invalid_request_uri_method"),
+    INVALID_TRANSACTION_DATA("invalid_transaction_data"),
+    WALLET_UNAVAILABLE("wallet_unavailable"),
+    WALLET_ERROR("wallet_error");
```
IngridPuppet, line 17: Where did you get all there errors from ? I assume some come from the spec... Others don't, such as `temporarily_unavailable`, I couldn't find it.

Overall, this enumeration was meant for errors that the plugin issues itself as it processes a response sent by the wallet. It starts being misleading to include every possible error sent by the wallet, and we certainly can't capture all. 

That's why I suggest to create an indirection. You could just add `WALLET_ERROR("wallet_error")` to signal that authorization failed because of an error sent by the wallet, then in the error description field in the context you concatenate what the wallet sent ("error: error_description"). I believe it's more reliable to proceed that way and avoid this extensive list.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/dto/ProcessingError.java
```diff
@@ -22,4 +37,12 @@ public enum ProcessingError {
     public String getErrorString() {
         return error;
     }
+
+    @JsonCreator
+    public static ProcessingError fromErrorString(String error) {
+        return Arrays.stream(values())
+                .filter(processingError -> processingError.error.equals(error))
+                .findFirst()
+                .orElse(WALLET_ERROR);
```
IngridPuppet: For the reasons stated above, `WALLET_ERROR` is definitely not a great default for `ProcessingError`. Also, this method should not be needed eventually because these processing errors are meant to be produced by the plugin, not provided by a third party.

IngridPuppet (changes requested):
I reviewed the changes and left some comments. Please could you check?

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointTest.java
forkimenjeckayang: Removed `shouldPersistWalletErrorResponse`  it duplicated the HAIP happy path (`shouldAcceptUnencryptedWalletErrorResponse`)

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/dto/ProcessingError.java
forkimenjeckayang: Refactored to your approach: only `WALLET_ERROR` for wallet-posted failures, with `error` / `error_description` concatenated in the context’s `error_description`. Plugin-issued values remain `auth_context_closed`, `invalid_vp_token`, and `vp_token_auth_error`. Thanks for catching this.

IngridPuppet (approved):
LGTM

PR #85, opened 2026-05-13
Enforce PKIX trust validation for Status List JWT x5c certificate chains

### LINKED ISSUE #31: Enforce trust in X5C certificate chain for Status List JWTs
The plugin currently uses the public key from the first certificate in the x5c header of a Status List JWT without validating the certificate chain. This allows self-signed or unauthorized certificates to be trusted, creating a security risk. The goal of this ticket is to add proper certificate chain validation.

Estimated time: 3days

Awambeng:
## Problem

The plugin previously trusted the public key from the first certificate in the `x5c` header of a Status List JWT without validating the certificate chain. This allowed self-signed or unauthorized certificates to be accepted, creating a security risk.

## Solution

This PR introduces proper PKIX-based certificate chain validation and trust enforcement for Status List JWT verification.

### Changes

- Added a truststore-based certificate validation flow for `x5c` certificate chains.
- Implemented PKIX certificate path validation utilities.
- Enhanced `TrustedStatusListJwtFetcher` to enforce certificate trust before accepting JWS signatures.
- Improved JWS signature validation and certificate validation logging for easier debugging and traceability.
- Updated test resources with trusted certificate chains and invalid-chain scenarios.
- Updated Testcontainer configuration to support truststore handling during integration tests.

## Result

Status List JWTs are now only trusted when their certificate chain validates against the configured truststore, preventing unauthorized or self-signed certificates from being accepted.

Closes #31


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/TrustedStatusListJwtFetcher.java
```diff
@@ -1,122 +1,171 @@
...
+        } catch (Exception e) {
+            throw new ReferencedTokenValidationException("Failed to create signature verifier for " + alg, e);
+        }
+    }
+
+    protected X509Certificate[] validateCertChain(List<String> x5c) throws ReferencedTokenValidationException {
+        TruststoreProvider truststoreProvider = session.getProvider(TruststoreProvider.class);
+        if (truststoreProvider == null || truststoreProvider.getTruststore() == null) {
+            logger.warn("No Keycloak global truststore configured. Falling back to default JVM truststore.");
+        }
```
forkimenjeckayang, line 128: This log says we fall back to the JVM truststore, but `PKIXVerificationUtil.validateChain` always fails closed when the truststore/roots are missing. Please remove this warn or change it to state that validation will fail without a configured truststore (same behaviour as x509 `CertificateValidator.validateTrust()`).

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/PKIXVerificationUtil.java
```diff
@@ -0,0 +1,88 @@
...
+            if (truststoreProvider == null
+                    || truststoreProvider.getRootCertificates().isEmpty()) {
+                throw new VerificationException("No trusted root certificates available for validation");
+            }
+
+            // Build PKIX parameters
+            Set<TrustAnchor> trustAnchors = truststoreProvider.getRootCertificates().values().stream()
+                    .flatMap(List::stream)
+                    .map(cert -> new TrustAnchor(cert, null))
+                    .collect(Collectors.toSet());
```
forkimenjeckayang, line 54: `TruststoreProvider.getIntermediateCertificates()` is never used — only roots become trust anchors. Keycloak’s x509 validator also puts intermediates (and the presented chain) into a PKIX `CertStore`. Can we add the same here so issuer CAs stored as intermediates in the truststore still work when they’re not all repeated in the JWT `x5c` array?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/TrustedStatusListJwtFetcher.java
```diff
@@ -1,122 +1,171 @@
...
 
-        return signatureProvider.verifier(keyWrapper);
+        try {
+            return PKIXVerificationUtil.validateChain(x5c, truststoreProvider);
+        } catch (VerificationException e) {
+            throw new ReferencedTokenValidationException(e.getMessage(), e.getCause());
```
forkimenjeckayang: Please preserve the full cause chain: use `new ReferencedTokenValidationException(e.getMessage(), e)` instead of `e.getCause()`. Otherwise the `VerificationException` from `PKIXVerificationUtil` is lost.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/TrustedStatusListJwtFetcher.java
```diff
@@ -1,122 +1,171 @@
...
+                if (Algorithm.ES512.equals(alg)) {
+                    javaAlg = "SHA512withECDSA";
+                }
+
+                Signature sig = Signature.getInstance(javaAlg);
+                sig.initVerify(publicKey);
+                sig.update(data);
+                if (!sig.verify(derSignature)) {
+                    throw new ReferencedTokenValidationException("Invalid JWS signature");
+                }
```
forkimenjeckayang: For EC we bypass `SignatureVerifierContext` and duplicate ECDSA/JWS logic.  Can we drop the custom EC branch and call `verifier.verify(data, signature)` for every `alg`?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/TrustedStatusListJwtFetcher.java
```diff
@@ -1,122 +1,171 @@
...
-            if (!verifier.verify(jws.getEncodedSignatureInput().getBytes(), jws.getSignature())) {
-                throw new ReferencedTokenValidationException("Invalid JWS signature");
-            }
-        } catch (Exception e) {
-            throw new ReferencedTokenValidationException("Error during JWS signature verification", e);
+    private static String algorithmToKeyType(String alg) {
+        if (Algorithm.ES256.equals(alg) || Algorithm.ES384.equals(alg) || Algorithm.ES512.equals(alg)) {
+            return KeyType.EC;
         }
+        return KeyType.RSA;
```
forkimenjeckayang: `algorithmToKeyType` now defaults unknown algorithms to RSA. Before we threw `Unsupported signature algorithm` for unknown `alg`. Any reason for this?

forkimenjeckayang (changes requested):
Few comments @Awambeng

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/TrustedStatusListJwtFetcher.java
Awambeng: Done. I updated the error message accordingly.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/PKIXVerificationUtil.java
Awambeng: Done.

Awambeng:
@forkimenjeckayang I’ve addressed the comments. Please take another look when you get the chance.

forkimenjeckayang (dismissed):
LGTM unless @IngridPuppet  has more comments

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/PKIXVerificationUtil.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public static X509Certificate[] validateChain(List<String> x5c, TruststoreProvider truststoreProvider)
+            throws VerificationException {
+        try {
+            if (x5c == null || x5c.isEmpty()) {
+                throw new VerificationException("Certificate chain is empty");
+            }
+
+            if (x5c.size() > 5) {
+                throw new VerificationException("Certificate chain too long: " + x5c.size());
```
IngridPuppet: Consider adding max length supported to the message.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/PKIXVerificationUtil.java
```diff
@@ -0,0 +1,101 @@
...
+
+            // Validate the path. The CertPath should not include the trust anchor itself.
+            List<X509Certificate> pathCerts = new ArrayList<>(certs);
+            X509Certificate lastCert = pathCerts.get(pathCerts.size() - 1);
+            boolean rootInPath = trustAnchors.stream()
+                    .anyMatch(anchor -> anchor.getTrustedCert().equals(lastCert));
+
+            if (rootInPath && pathCerts.size() > 1) {
+                pathCerts.remove(pathCerts.size() - 1);
+            }
```
IngridPuppet, line 98: What happens if there is only one certificate in X5C and that one is trusted, i.e., part of trust anchors? Seeking clarity because that's how test envs will commonly set up trust in a self-signed certificate. It seems this implementation will reject self-signed certs despite them added to the trust store, which I believe is not wanted.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/PKIXVerificationUtil.java
```diff
@@ -0,0 +1,101 @@
...
+
+            return certs.toArray(new X509Certificate[0]);
+
+        } catch (VerificationException e) {
+            throw e;
+        } catch (GeneralSecurityException e) {
+            throw new VerificationException("Certificate chain validation failed: " + e.getMessage(), e);
+        } catch (Exception e) {
+            throw new VerificationException("Error during certificate chain validation: " + e.getMessage(), e);
+        }
```
IngridPuppet: Could be treated under the same catch. Also, no need to explicitly append `e.getMessage()`, it will appear in the stack strace.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/PKIXVerificationUtil.java
```diff
@@ -0,0 +1,101 @@
...
+            // For production-grade revocation, set to true and configure a PKIXRevocationChecker or
+            // provide a CRL cert store via params.addCertStore().
+            params.setRevocationEnabled(false);
+            // Sync with Keycloak offset time for testing (and production time consistency)
+            params.setDate(new Date(Time.currentTimeMillis()));
+
+            // Add intermediate certificates from the truststore and the presented chain
+            // as a CertStore so PKIX can bridge chains where the issuer CA is stored as
+            // an intermediate (not repeated in the JWT x5c array).
+            Collection<X509Certificate> allIntermediates = new ArrayList<>(certs);
```
IngridPuppet: Certs in the presented chain will be part of the cert path. What's the value of adding them to this cert store?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/SimpleStatusListJwtFetcher.java
```diff
@@ -25,6 +25,10 @@ public SimpleStatusListJwtFetcher(KeycloakSession session) {
 
     @Override
     public String fetchStatusListJwt(String uri) throws ReferencedTokenValidationException {
+        return _fetchStatusListJwt(uri);
+    }
+
+    protected String _fetchStatusListJwt(String uri) throws ReferencedTokenValidationException {
```
IngridPuppet: What's the added value of this change? Java makes it possible to call "super.fetchStatusListJwt" in child classes.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/TrustedStatusListJwtFetcher.java
```diff
@@ -1,122 +1,157 @@
...
-        // Extract verifying key from X5C cert chain
-        // TODO: Enforce trust in X5C chain
-        SignatureVerifierContext verifier = getVerifierFromX5C(jws);
-
-        // Verify signature
-        validateJwsSignature(jws, verifier);
+        verifyStatusListJwt(jws, uri);
 
         return statusListJwt;
     }
```
IngridPuppet, line 48: Could you restore comments?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/TrustedStatusListJwtFetcher.java
```diff
@@ -1,122 +1,157 @@
...
         }
+    }
 
-        KeyWrapper keyWrapper = new KeyWrapper();
-        keyWrapper.setPublicKey(cert.getPublicKey());
-        keyWrapper.setType(algorithmToKeyType(alg));
-        keyWrapper.setAlgorithm(alg);
+    protected X509Certificate[] validateCertChain(List<String> x5c) throws ReferencedTokenValidationException {
+        TruststoreProvider truststoreProvider = session.getProvider(TruststoreProvider.class);
+        if (truststoreProvider == null || truststoreProvider.getTruststore() == null) {
```
IngridPuppet, line 126: Is there a native mechanism to have a truststore just for status lists? Maybe we will not want to extend trust to absolutely all features in Keycloak.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/TrustedStatusListJwtFetcherTest.java
```diff
@@ -61,26 +101,190 @@ public void shouldRejectNonHttpsURIs() {
...
+        setupTrustForStatusListJwt();
+
+        TrustedStatusListJwtFetcher spyFetcher = Mockito.spy(fetcher);
+        Mockito.doThrow(new VerificationException("Leaf certificate must not be a CA"))
+                .when(spyFetcher)
+                .validateLeafCertificate(Mockito.any());
+
+        var e = assertThrows(ReferencedTokenValidationException.class, () -> spyFetcher.fetchStatusListJwt(uri));
+        assertEquals("Leaf certificate must not be a CA", e.getCause().getMessage());
+    }
```
IngridPuppet, line 206: You should prepare a sample with a leaf as CA. if you mock `validateLeafCertificate`, then you're not testing it.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/TrustedStatusListJwtFetcherTest.java
```diff
@@ -61,26 +101,190 @@ public void shouldRejectNonHttpsURIs() {
...
+        TrustedStatusListJwtFetcher spyFetcher = Mockito.spy(fetcher);
+        Mockito.doThrow(new VerificationException("Leaf certificate missing Digital Signature KeyUsage"))
+                .when(spyFetcher)
+                .validateLeafCertificate(Mockito.any());
+
+        var e = assertThrows(ReferencedTokenValidationException.class, () -> spyFetcher.fetchStatusListJwt(uri));
+        assertEquals(
+                "Leaf certificate missing Digital Signature KeyUsage",
+                e.getCause().getMessage());
+    }
```
IngridPuppet, line 238: You should prepare a sample with a missing KU. if you mock `validateLeafCertificate`, then you're not testing it. There should be online tools to help you generate these cert samples.

IngridPuppet (changes requested):
I reviewed the changes and left some comments. Please could you check?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/PKIXVerificationUtil.java
Awambeng: Added a `MAX_CHAIN_LENGTH = 5` constant and updated the error message to include the limit

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/PKIXVerificationUtil.java
Awambeng: I added an explicit check to support self-signed trusted certificates. If the leaf certificate itself is found in the trust anchors, it is validated for expiration and accepted immediately, bypassing the CertPathValidator which typically expects a multi-cert path.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/PKIXVerificationUtil.java
Awambeng: Combined the `GeneralSecurityException` and general `Exception` catch blocks into a single block that throws a simplified `VerificationException`, removing redundant stack trace logic from the message.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/PKIXVerificationUtil.java
Awambeng: Optimized the CertStore population. It now only includes intermediate certificates from the Keycloak truststore. The certificates from the presented x5c chain are correctly omitted as they are already part of the CertPath

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/SimpleStatusListJwtFetcher.java
Awambeng: Refactored the fetcher hierarchy to use a protected `fetchStatusListFromUri` method. This allows subclasses (and test stubs) to override just the retrieval logic while keeping the validation flow (like HTTPS enforcement) centralized in the main `fetchStatusListJwt` entry point.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/TrustedStatusListJwtFetcher.java
Awambeng: Restored the inline comments

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/TrustedStatusListJwtFetcherTest.java
Awambeng: Hardened the test suite by removing `Mockito.spy` on internal validation methods. I’ve added a utility to generate real X.509 certificates with specific "bad" attributes (e.g., BasicConstraints: CA=true or missing KeyUsage). These are used to build actual JWS objects for the tests, ensuring the real validation logic is exercised instead of just mocked.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/tokenstatus/http/TrustedStatusListJwtFetcher.java
Awambeng: Keycloak does not provide a native mechanism to isolate trust specifically for status lists, which is why I opted for the global TruststoreProvider SPI.

IngridPuppet (approved):

IngridPuppet:
Thank you for the work on the PR.

PR #86, opened 2026-05-15
Implement new response format

### LINKED ISSUE #64: Migrate Authorization Response Parsing to vp_token Structure
Description: Stop requiring presentation_submission; parse vp_token JSON object keyed by query IDs and extract matching SD-JWT VP candidates.
Specification reference: https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-response-parameters
Recommended approach: Implement parser supporting final structure first.

ETA: 1 day

Ogenbertrand:
closes: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/64

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ResponseObject.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ResponseObject.java
@@ -3,7 +3,10 @@ package io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JsonNode;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import org.keycloak.util.JsonSerialization;
@@ -22,7 +25,6 @@ public class ResponseObject {
     public static final String VP_TOKEN_KEY = "vp_token";
     public static final String STATE_KEY = "state";
 
-    @JsonProperty(VP_TOKEN_KEY)
     private Map<String, List<String>> vpToken;
 
     @JsonProperty(STATE_KEY)
@@ -31,25 +33,52 @@ public class ResponseObject {
     ResponseObject() {}
 
     public ResponseObject(String vpToken, String state) throws JsonProcessingException {
-        this.vpToken = parseVpToken(vpToken);
+        this.vpToken = readVpToken(vpToken);
         this.state = state;
     }
 
-    private static Map<String, List<String>> parseVpToken(String vpToken) throws JsonProcessingException {
+    private static Map<String, List<String>> readVpToken(String vpToken) throws JsonProcessingException {
         if (StringUtil.isBlank(vpToken)) {
             throw new IllegalArgumentException("vp_token must not be null or blank");
         }
 
-        return JsonSerialization.mapper.readValue(vpToken, new TypeReference<Map<String, List<String>>>() {});
+        return parseVpToken(JsonSerialization.mapper.readTree(vpToken));
     }
 
+    private static Map<String, List<String>> parseVpToken(JsonNode vpToken) {
+        if (vpToken == null || !vpToken.isObject()) {
+            throw new IllegalArgumentException("vp_token must be a JSON object keyed by DCQL credential query IDs");
+        }
+
+        Map<String, List<String>> result = new LinkedHashMap<>();
+        vpToken.properties().forEach(entry -> {
+            JsonNode presentations = entry.getValue();
+            if (!presentations.isArray()) {
+                throw new IllegalArgumentException(
+                        "vp_token entry `%s` must be an array of presentations".formatted(entry.getKey()));
+            }
+
+            List<String> presentationStrings = new ArrayList<>();
+            presentations.forEach(presentation -> {
+                if (!presentation.isTextual()) {
+                    throw new IllegalArgumentException(
+                            "vp_token entry `%s` must contain string presentations".formatted(entry.getKey()));
+                }
+                presentationStrings.add(presentation.asText());
+            });
+            result.put(entry.getKey(), List.copyOf(presentationStrings));
+        });
+        return Collections.unmodifiableMap(result);
+    }
+
+    @JsonProperty(VP_TOKEN_KEY)
     public Map<String, List<String>> getVpToken() {
         return vpToken;
     }
 
-    public ResponseObject setVpToken(Map<String, List<String>> vpToken) {
-        this.vpToken = vpToken;
-        return this;
+    @JsonProperty(VP_TOKEN_KEY)
+    public void setVpToken(JsonNode vpToken) {
+        this.vpToken = parseVpToken(vpToken);
     }
 
     public String getState() {
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ResponseObject.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ResponseObject.java
@@ -25,7 +25,7 @@ public class ResponseObject {
     public static final String VP_TOKEN_KEY = "vp_token";
     public static final String STATE_KEY = "state";
 
-    private Map<String, List<JsonNode>> vpToken;
+    private Map<String, List<String>> vpToken;
 
     @JsonProperty(STATE_KEY)
     private String state;
@@ -37,7 +37,7 @@ public class ResponseObject {
         this.state = state;
     }
 
-    private static Map<String, List<JsonNode>> readVpToken(String vpToken) throws JsonProcessingException {
+    private static Map<String, List<String>> readVpToken(String vpToken) throws JsonProcessingException {
         if (StringUtil.isBlank(vpToken)) {
             throw new IllegalArgumentException("vp_token must not be null or blank");
         }
@@ -45,12 +45,12 @@ public class ResponseObject {
         return parseVpToken(JsonSerialization.mapper.readTree(vpToken));
     }
 
-    private static Map<String, List<JsonNode>> parseVpToken(JsonNode vpToken) {
+    private static Map<String, List<String>> parseVpToken(JsonNode vpToken) {
         if (vpToken == null || !vpToken.isObject()) {
             throw new IllegalArgumentException("vp_token must be a JSON object keyed by DCQL credential query IDs");
         }
 
-        Map<String, List<JsonNode>> result = new LinkedHashMap<>();
+        Map<String, List<String>> result = new LinkedHashMap<>();
         vpToken.properties().forEach(entry -> {
             JsonNode presentations = entry.getValue();
             if (!presentations.isArray()) {
@@ -58,13 +58,13 @@ public class ResponseObject {
                         "vp_token entry `%s` must be an array of presentations".formatted(entry.getKey()));
             }
 
-            List<JsonNode> presentationNodes = new ArrayList<>();
+            List<String> presentationNodes = new ArrayList<>();
             presentations.forEach(presentation -> {
                 if (!presentation.isTextual() && !presentation.isObject()) {
                     throw new IllegalArgumentException("vp_token entry `%s` must contain string or object presentations"
                             .formatted(entry.getKey()));
                 }
-                presentationNodes.add(presentation.deepCopy());
+                presentationNodes.add(presentation.isTextual() ? presentation.asText() : presentation.toString());
             });
             result.put(entry.getKey(), List.copyOf(presentationNodes));
         });
@@ -72,7 +72,7 @@ public class ResponseObject {
     }
 
     @JsonProperty(VP_TOKEN_KEY)
-    public Map<String, List<JsonNode>> getVpToken() {
+    public Map<String, List<String>> getVpToken() {
         return vpToken;
     }
 
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ResponseObject.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ResponseObject.java
@@ -25,7 +25,7 @@ public class ResponseObject {
     public static final String VP_TOKEN_KEY = "vp_token";
     public static final String STATE_KEY = "state";
 
-    private Map<String, List<String>> vpToken;
+    private Map<String, List<JsonNode>> vpToken;
 
     @JsonProperty(STATE_KEY)
     private String state;
@@ -37,7 +37,7 @@ public class ResponseObject {
         this.state = state;
     }
 
-    private static Map<String, List<String>> readVpToken(String vpToken) throws JsonProcessingException {
+    private static Map<String, List<JsonNode>> readVpToken(String vpToken) throws JsonProcessingException {
         if (StringUtil.isBlank(vpToken)) {
             throw new IllegalArgumentException("vp_token must not be null or blank");
         }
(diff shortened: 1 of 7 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -176,23 +160,21 @@ private String extractSdJwtVpTokenWithDCQL(
                     store);
```
forkimenjeckayang: This enforces exactly one SD-JWT after extraction. That matches SdJwtAuthenticator (single auth note) and current ITs, but VpTokenCandidateExtractor already accepts multiple strings when DCQL multiple is true. A wallet sending {"id":["vp1","vp2"]} would pass extraction then fail here. Please document "login uses one SD-JWT only" or align extractor + auth (e.g. reject multiple at extract time for this flow, or take the first).

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VpTokenCandidateExtractor.java
```diff
@@ -0,0 +1,167 @@
...
+        validatePresentationArrays(vpToken, credentialsById);
+        validateRequiredCredentialSets(dcqlQuery, vpToken.keySet(), credentialsById);
+
+        List<String> candidates = new ArrayList<>();
+        vpToken.forEach((queryId, presentations) -> {
+            Credential credential = credentialsById.get(queryId);
+            // This authenticator currently handles SD-JWT VCs only; other formats are left out deliberately.
+            if (VCFormat.SD_JWT_VC.equals(credential.getFormat())) {
+                candidates.addAll(presentations);
+            }
```
forkimenjeckayang, line 44: Only VCFormat.SD_JWT_VC presentations are collected; other formats in the map are skipped. If the wallet returns only mdoc (or other) keys, extraction fails with no SD-JWT presentations. That is fine for this plugin scope, but if vp_token can include mixed formats, consider a clearer error when known query ids are present but none are SD-JWT (instead of a generic no SD-JWT message).

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ResponseObject.java
```diff
@@ -31,25 +33,52 @@ public class ResponseObject {
...
+            presentations.forEach(presentation -> {
+                if (!presentation.isTextual()) {
+                    throw new IllegalArgumentException(
+                            "vp_token entry `%s` must contain string presentations".formatted(entry.getKey()));
+                }
+                presentationStrings.add(presentation.asText());
+            });
+            result.put(entry.getKey(), List.copyOf(presentationStrings));
+        });
+        return Collections.unmodifiableMap(result);
```
forkimenjeckayang, line 71: `parseVpToken` accepts {} (empty object). `VpTokenCandidateExtractor` then rejects with `vp_token must be a non-empty JSON object.` Consider rejecting an empty object here with the same message so form-param / JWE parsing fails fast with a consistent error. Also Final spec allows presentation values as string or object depending on format. This parser requires textual array elements, which matches SD-JWT VP strings for this plugin. OK if documented as SD-JWT-only; wallets sending JSON object presentations for other formats will get a clear parse error.

forkimenjeckayang (changes requested):
few comments @Ogenbertrand

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VpTokenCandidateExtractor.java
Ogenbertrand: Please check the new changes.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ResponseObject.java
Ogenbertrand: Please check the new changes.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ResponseObject.java
```diff
@@ -22,34 +25,60 @@ public class ResponseObject {
...
 
-        return JsonSerialization.mapper.readValue(vpToken, new TypeReference<Map<String, List<String>>>() {});
+        return parseVpToken(JsonSerialization.mapper.readTree(vpToken));
     }
 
-    public Map<String, List<String>> getVpToken() {
+    private static Map<String, List<JsonNode>> parseVpToken(JsonNode vpToken) {
+        if (vpToken == null || !vpToken.isObject()) {
+            throw new IllegalArgumentException("vp_token must be a JSON object keyed by DCQL credential query IDs");
+        }
```
forkimenjeckayang, line 51: `parseVpToken` allows `{}`. Please reject `vpToken.isEmpty()` on this path (and align OAuth error code with extractor if you want `invalid_vp_token`). Add test in `VpTokenCandidateExtractorTest` or via `new ResponseObject("{}", "state")`.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ResponseObject.java
```diff
@@ -22,34 +25,60 @@ public class ResponseObject {
...
+
+        Map<String, List<JsonNode>> result = new LinkedHashMap<>();
+        vpToken.properties().forEach(entry -> {
+            JsonNode presentations = entry.getValue();
+            if (!presentations.isArray()) {
+                throw new IllegalArgumentException(
+                        "vp_token entry `%s` must be an array of presentations".formatted(entry.getKey()));
+            }
```
forkimenjeckayang, line 59: Optional: reject empty presentation arrays here for fail-fast, or leave to `VpTokenCandidateExtractor.validatePresentationArrays`.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VpTokenCandidateExtractor.java
```diff
@@ -0,0 +1,194 @@
...
+            Credential credential = credentialsById.get(queryId);
+            // This authenticator currently handles SD-JWT VCs only; other formats are left out deliberately.
+            if (VCFormat.SD_JWT_VC.equals(credential.getFormat())) {
+                candidates.addAll(extractSdJwtStrings(queryId, presentations));
+            }
+        });
+
+        if (candidates.isEmpty()) {
+            throw invalid("Presented vp_token map does not contain any SD-JWT VP presentations");
+        }
```
forkimenjeckayang, line 49: When `candidates.isEmpty()` but `vpToken` has keys, consider listing presented credential formats (e.g. mdoc-only) instead of only "does not contain any SD-JWT VP presentations".

forkimenjeckayang (changes requested):
additional comments @Ogenbertrand

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ResponseObject.java
Ogenbertrand: This test `shouldRejectNonObjectVpToken` handles that  already

Ogenbertrand: This test `shouldRejectNonObjectVpToken` handles that already.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ResponseObject.java
Ogenbertrand: This available check is enough IMO.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VpTokenCandidateExtractor.java
Ogenbertrand: IMO i don't think this is really neccessary tight what is the essence of listing formats available if the candidate is empty, i think updating the error message should be enough. WDYT ?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ResponseObject.java
```diff
@@ -22,34 +25,60 @@ public class ResponseObject {
     public static final String VP_TOKEN_KEY = "vp_token";
     public static final String STATE_KEY = "state";
 
-    @JsonProperty(VP_TOKEN_KEY)
-    private Map<String, List<String>> vpToken;
+    private Map<String, List<JsonNode>> vpToken;
```
IngridPuppet, line 29: Why do you need to change String to JsonNode and make your life harder everywhere? Any plan to accept a presentation format that's not a String?

Ogenbertrand: Thanks for pointing this out, i didn't realize it.
Leme revert that please.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ResponseObject.java
```diff
@@ -31,25 +33,52 @@ public class ResponseObject {
...
+                if (!presentation.isTextual() && !presentation.isObject()) {
+                    throw new IllegalArgumentException("vp_token entry `%s` must contain string or object presentations"
+                            .formatted(entry.getKey()));
+                }
+                presentationNodes.add(presentation.isTextual() ? presentation.asText() : presentation.toString());
+            });
+            result.put(entry.getKey(), List.copyOf(presentationNodes));
+        });
+        return Collections.unmodifiableMap(result);
+    }
```
IngridPuppet, line 72: Could you clarify why you need this manual parsing instead of the previous logic?

```java
        return JsonSerialization.mapper.readValue(vpToken, new TypeReference<Map<String, List<String>>>() {});
```

I wonder why any change is needed in this class.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -130,43 +132,28 @@ private static String getAuthenticatorErrorMessage(Response response) {
...
-            String detailed = "Could not parse SD-JWT VP token contained in `vp_token`";
+            parsedVpToken = vpTokenCandidateExtractor.extractSingleSdJwtCandidate(
+                    authContext.getRequestObject().getDcqlQuery(), responseObject.getVpToken());
+        } catch (VpTokenCandidateExtractor.InvalidVpTokenException e) {
             throw failWithHttpException(
                     ProcessingError.INVALID_VP_TOKEN,
                     "Invalid vp_token",
-                    detailed,
+                    e.getMessage(),
                     Response.Status.BAD_REQUEST,
```
IngridPuppet, line 145: It's important to log the error with the exception cause, so a stack trace is added to the logs.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VpTokenCandidateExtractor.java
```diff
@@ -0,0 +1,188 @@
...
+        if (!unknownOptionIds.isEmpty()) {
+            throw new IllegalStateException(
+                    "Invalid DCQL query in authorization context. Credential set references unknown credential IDs: "
+                            + unknownOptionIds);
+        }
+    }
+
+    private InvalidVpTokenException invalid(String message) {
+        return new InvalidVpTokenException(message);
+    }
```
IngridPuppet, line 180: I'd rather you just use `new InvalidVpTokenException` directly for clarity.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VpTokenCandidateExtractor.java
IngridPuppet, line 1: This PR globally sends a lot of changes but I fail to get what issue it attempts to solve. Please, could you clarify the following?
- What's that the plugin couldn't achieve previously or that was non-compliant with the spec?
- How do these changes bridge that gap and what's the added value?

IngridPuppet (changes requested):
I'm really confused about what this PR attempts to achieve. Could you check my concerns below so I could better review the PR?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/ResponseObject.java
Ogenbertrand: `vp_token` is now parsed as a JSON object keyed by DCQL credential query IDs, values must be arrays, and SD-JWT VP candidates are extracted by matching those keys against the DCQL query. See it here: https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-response-parameters.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VpTokenCandidateExtractor.java
Ogenbertrand: In draft 20, `vp_token` was a flat container that returned one Verifiable Presentation or an array of presentations without explicitly linking them to requested credential queries, whereas in the final version1_0, vp_token became a DCQL-query-keyed JSON object where each entry explicitly maps a credential query ID to the presentations that satisfy that specific query. Those were the gaps i addressed.

Ogenbertrand: Also in the final version, the multiple parameter became a formal DCQL constraint that determines whether a credential query may return more than one presentation.

IngridPuppet: > vp_token became a DCQL-query-keyed JSON object where each entry explicitly maps a credential query ID to the presentations that satisfy that specific query

Recall we've already started implementing 1.0-Final. If I'm not mistaken, this series of tickets aim to address what was not covered yet. This is what we had at this point:

```java
private Map<String, List<String>> vpToken;
```

This is 1.0-Final structure not draft20 (which used `Object` previously). Do you see any reason why this would not be sufficient to support vp tokens as "a DCQL-query-keyed JSON object where each entry explicitly maps a credential query ID to the presentations that satisfy that specific query"?

> Also in the final version, the multiple parameter became a formal DCQL constraint that determines whether a credential query may return more than one presentation.

Yes, this is something new. But adding the field to the DCQL model and not doing anything with it is really not changing anything concretely. The plugin creates DCQL requests; it does not consume external requests. Are there any clear plans about how to process responses with multiple presentations for the same credential query ID? This is a complex topic on its own, that IMO should not be addressed in a PR that's just titled "Implement new response format".

Overall, like I said previously, the PR makes changes here and there, but I really don't see what's the new feature, what's the problem addressed, etc. Even with the point on parameter `multiple` in DCQL model, we still only extract just one presentation like previously. Do you think there some key aspect of the PR I fail to identify?

Ogenbertrand: Do you suggest we close this PR and open a follow up PR for the `multiple` parameter ?

IngridPuppet: > open a follow up PR for the `multiple` parameter

This definitely. We'll then discuss the need for it separately and agree on a plan before implementation.

> Do you suggest we close this PR

If all other code is proven redundant, then yes.

Ogenbertrand: I'll therefore proceed by closing this PR.
Thanks for the Insights @IngridPuppet.

PR #90, opened 2026-05-19
feat(oid4vp): verifier-side vp_token validation pipeline (#76)

### LINKED ISSUE #76: Verifier-side validation pipeline
Description: Independently verify every returned presentation: integrity, replay binding (nonce, audience / domain), format-specific checks, trust/revocation as required, and that results satisfy the issued DCQL. Do not assume the wallet enforced DCQL.
Spec: Final 1.0 §8.6, §14.9
https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-vp-token-validation
https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-security-checks-on-the-retu
Approach: Structure → per-format → DCQL satisfaction order.

EST: 1 day

mbunwe-victor:
## Summary

Implements verifier-side `vp_token` validation for OpenID4VP user authentication ([#76](https://github.com/.../issues/76)).

The verifier independently validates every returned presentation before authentication: it does not assume the wallet enforced DCQL (OID4VP §14.9). Validation runs in `AuthorizationResponseService` before the SD-JWT authenticator.

**Pipeline order (§8.6):** structure → per-format → DCQL satisfaction

---

### Validation pipeline (`validation` package)

- **`VpTokenValidationPipeline`** — orchestrates the three phases
- **`VpTokenStructureValidator`** — `vp_token` shape vs issued DCQL (`credential_sets`, `multiple`, unknown ids, etc.)
- **`SdJwtPresentationValidator`** — format-specific checks for `dc+sd-jwt` / `SD_JWT_VC`:
  - integrity / authenticity (`SdJwtPresentationConsumer`, `SelfTrustedSdJwtIssuer`)
  - holder binding (nonce + audience)
  - optional revocation (`ReferencedTokenValidator` when configured)
- **`DcqlSatisfactionValidator`** — verifier-side DCQL satisfaction (`meta.vct_values`, `claims`, `claim_sets`, typed `values`)
- **`HolderBindingAudienceResolver`** — audience from `client_id` or `origin:` derived from `response_uri`
- **`VpTokenValidationService`** — builds context from the stored authorization request and runs the pipeline

### Integration

- **`AuthorizationResponseService`** — runs validation on authorization response; maps failures to `INVALID_VP_TOKEN` or `VP_TOKEN_AUTH_ERROR`
- **`SdJwtAuthenticator`** — skips duplicate crypto validation when `vp_token_validated` is set; focuses on user resolution

### Model

- **`Claim.values`** — `List<Object>` so DCQL `values` can be strings, integers, or booleans (§6.3), with typed matching in `DcqlClaimValues`

---

## Spec References

- [§8.6 VP Token Validation](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-vp-token-validation)
- [§14.9 Security Checks on the Returned Credentials and Presentations](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-security-checks-on-the-retu)

---

Closes: #76


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthenticator.java
```diff
@@ -65,33 +71,38 @@ public void authenticate(AuthenticationFlowContext context) {
...
+                    tokenStatusValidator.validate(sdJwt.getIssuerSignedJWT().getPayload());
+                } catch (ReferencedTokenValidationException e) {
+                    logger.errorf(e, "Token status verification failed (authSession = %s)", correlationId(context));
+                    failRejectingPresentedSdJwtToken(context, "Token status verification failed", e);
+                    return;
+                }
+            }
+        } else {
+            logger.debug("Skipping duplicate SD-JWT validation; vp_token pipeline already validated presentation");
         }
```
forkimenjeckayang: After moving validation into `VpTokenValidationPipeline`, consider failing closed when `VP_TOKEN_VALIDATED_KEY` is not set (for this flow), instead of re-running a different, legacy validation path. The fallback uses `getPresentationRequirements()` and no longer receives `nonce`/`aud` auth notes from `AuthorizationResponseService`.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/validation/VpTokenValidationService.java
```diff
@@ -0,0 +1,79 @@
...
+            Class<?> mockFetcherClass = Class.forName(
+                    "io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.stub.CustomSdJwtAuthenticatorFactory$MockTrustedStatusListJwtFetcher");
+            Constructor<?> constructor = mockFetcherClass.getConstructor(KeycloakSession.class);
+            return (StatusListJwtFetcher) constructor.newInstance(session);
+        } catch (ClassNotFoundException ignored) {
+            return new SdJwtAuthenticatorFactory().createStatusListJwtFetcher(session);
+        } catch (ReflectiveOperationException e) {
+            throw new IllegalStateException("Failed to initialize status list JWT fetcher", e);
+        }
+    }
```
forkimenjeckayang: Can we resolve `StatusListJwtFetcher` the same way as `SdJwtAuthenticatorFactory.create()` — via the registered `AuthenticatorFactory` SPI — instead of `Class.forName`? That keeps test doubles (`CustomSdJwtAuthenticatorFactory`) and production behavior aligned without reflection.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthRequirements.java
```diff
@@ -103,6 +106,47 @@ public PresentationRequirements getPresentationRequirements() {
...
+
+        if (credentialQuery.getClaims() != null) {
+            for (Claim claim : credentialQuery.getClaims()) {
+                if (claim.getPath() == null || claim.getPath().isEmpty()) {
+                    continue;
+                }
+                String claimName = claim.getPath().getLast();
+                requirements.addClaimRequirement(claimName, ".*");
+            }
+        }
```
forkimenjeckayang: `getPresentationRequirementsForCredential()` uses only the last path segment for `SimplePresentationDefinition`, but DCQL satisfaction uses full §7.1 paths. Please align these (e.g. derive presentation requirements from the same path resolution used in `SdJwtClaimReader`) so §8.6 format checks and §6.4 DCQL checks cannot diverge.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/validation/VpTokenPresentationDecoder.java
```diff
@@ -0,0 +1,21 @@
...
+    private VpTokenPresentationDecoder() {}
+
+    static String decodeIfBase64Url(String input) {
+        try {
+            byte[] decoded = Base64.getUrlDecoder().decode(input);
+            return new String(decoded, StandardCharsets.UTF_8);
+        } catch (IllegalArgumentException e) {
+            return input;
+        }
+    }
```
forkimenjeckayang, line 66: Please document when wallets are expected to return raw SD-JWT vs base64url-wrapped SD-JWT, or detect format explicitly (e.g. JWT shape `x.y.z`) before attempting base64url decode.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/validation/VpTokenValidationResult.java
```diff
@@ -0,0 +1,30 @@
...
+        return presentations.getFirst();
+    }
+
+    /**
+     * @deprecated use {@link #requireSinglePresentation()}
+     */
+    @Deprecated
+    public PresentedCredential singlePresentation() {
+        return requireSinglePresentation();
+    }
```
forkimenjeckayang: `VpTokenValidationResult` is a new file added by this PR, yet it already declares a `@Deprecated` `singlePresentation()` method that just delegates to `requireSinglePresentation()`. `rg singlePresentation\b src/` returns no callers anywhere on the branch. 
Shipping a brand-new API with a deprecated alias that has zero callers is dead code on day one. It also primes future contributors to keep it forever.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointBase.java
forkimenjeckayang, line 88: The pre-existing instance method `getSdjwtAuthenticatorConfig()` previously called `getOid4vpAuthFlow()`, which **throws** `IllegalStateException("Authentication flow '%s' not found. Such is supposed to be built-in")` when the flow is missing. The new `resolveSdJwtAuthenticatorConfig(KeycloakSession session)` silently returns `new AuthenticatorConfigModel()` instead. `getSdjwtAuthenticatorConfig()` was refactored to delegate to the new static, inheriting the more lenient behavior.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/validation/DcqlClaimValues.java
```diff
@@ -0,0 +1,64 @@
...
+            return expected.equalsIgnoreCase(String.valueOf(actual.asBoolean()));
+        }
+        if (actual.isIntegralNumber()) {
+            return expected.equals(String.valueOf(actual.asLong()));
+        }
+        if (actual.isNumber()) {
+            return expected.equals(String.valueOf(actual.asDouble()));
+        }
+        return expected.equals(actual.toString());
+    }
```
forkimenjeckayang, line 55: A DCQL constraint of `"values": ["true"]` would unexpectedly accept the boolean `true` claim, and `"values": ["42"]` would accept an integer `42`. This is unlikely to cause a security issue on its own but reduces the determinism of `values` constraints. Either restrict the matcher to strict JSON-type equality, or document the coercion explicitly in the class Javadoc. The dispatch by `expected`'s Java type in `equals(...)` already separates string/boolean/integer/number cases — collapsing `equalsString`'s non-textual branches to `return false` would suffice.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -69,21 +71,33 @@ public void processAuthorizationResponse(
...
+            throw failWithHttpException(
+                    formatFailure ? ProcessingError.VP_TOKEN_AUTH_ERROR : ProcessingError.INVALID_VP_TOKEN,
+                    formatFailure ? "Invalid SD-JWT presentation" : "Invalid vp_token",
+                    detailedMessage,
+                    formatFailure ? Response.Status.UNAUTHORIZED : Response.Status.BAD_REQUEST,
+                    authContext,
+                    store);
+        }
+
+        String sdJwtVp = validationResult.requireSinglePresentation().encodedPresentation();
```
forkimenjeckayang: `VpTokenValidationPipeline` can return more than one `PresentedCredential` when (a) several credential queries appear in `vp_token` (e.g. satisfied `credential_sets`), or (b) `Credential.multiple` is true and the wallet returns multiple presentation strings for one id (structure allows this; the pipeline appends one `PresentedCredential` per string in `VpTokenValidationPipeline` lines 50–58).

This call uses `requireSinglePresentation()`, which throws `IllegalStateException` if `presentations.size() != 1`. That exception is not handled by the `VpTokenValidationException` block above (lines 75–90), so a successful multi-presentation response would likely surface as an internal error rather than a controlled 4xx.

The default OID4VP login flow (single credential, single presentation) is unaffected, but this is a latent gap once `multiple: true` or multi-credential queries are used in production—the old `extractSdJwtVpTokenWithDCQL` “exactly one token” guard was removed without replacing it here.

Please either document/reject multi-presentation success for this login flow, or handle `presentations.size() > 1` explicitly (and add an integration test).

forkimenjeckayang (changes requested):
A few comments in addition to this

- New `validation` package is a good boundary, but `VpTokenValidationService` still reaches into `SdJwtAuthenticatorFactory` and duplicates authenticator config resolution (`OID4VPUserAuthEndpointBase.resolveSdJwtAuthenticatorConfig`). Consider a small factory/provider (e.g. `VpTokenValidationService.create(session)`) registered alongside other OID4VP services for consistency with Keycloak SPI style.

- `src/test/.../validation/*Test.java` (unit only): Strong unit coverage for pipeline components; no new end-to-end test proving `AuthorizationResponseService` sets `VP_TOKEN_VALIDATED_KEY` and rejects bad `vp_token` before `authenticateOnly()`. One HAIP/base endpoint test with an intentionally wrong `vp_token` map (wrong DCQL id / missing claim) would lock the wiring against regressions.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/validation/ClaimsPathProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+        return next;
+    }
+
+    /**
+     * Converts a DCQL path of strings to the mixed pointer components used by §7.1.
+     */
+    public static List<Object> toPathComponents(List<String> path) {
+        return List.copyOf(path);
+    }
+
```
forkimenjeckayang, line 116: Currently, this utility method directly converts a `List<String>` to a `List<Object>` via `List.copyOf(path)` because our flat-path schemas only use string-based properties. 

However, DCQL §7.1 allows claims path pointers to be a mixed array containing strings (object keys), integers (array indices), or `null` (all array elements) — for example `["address", null, "street_address"]`.

**Suggestion:**
For future-proofing, please document or leave a note that if `Claim.java` path parsing is ever widened to support mixed JSON-pointer types (such as integers or nulls), this mapper should be updated to parse numeric indices or null values from the path array accordingly.

forkimenjeckayang (changes requested):

IngridPuppet (changes requested):
This PR is excessively bloated and, in my view, unsafe to merge. Could you split it into two to three smaller PRs with a specific focus? 

So many changes look unwanted. For example, why is nonce/aud replay protection taken out of SdJwtAuthenticator? Why are we abandoning the SD-JWT verification infrastructure in upstream Keycloak, represented by `SdJwtPresentationConsumer`?

Is the PR concerned about DCQL matching, SD-JWT validation, or Holder binding enforcement?

In summary, please break down the intent of these changes into multiple isolated concerns, then send independent PRs after some discussion if required.

mbunwe-victor:
Hello @IngridPuppet, below are the splitted PR as you requested  
PR 1: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/94
PR 2: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/95
PR 3: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/96

IngridPuppet:
> Hello @IngridPuppet, below are the splitted PR as you requested

Hello @mbunwe-victor - I acknowledge the creation of the new PRs. Any reason for keeping this PR open? The new PRs can (and should) be tracked in ticket https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/76.

mbunwe-victor:
> > Hello @IngridPuppet, below are the splitted PR as you requested
> 
> Hello @mbunwe-victor - I acknowledge the creation of the new PRs. Any reason for keeping this PR open? The new PRs can (and should) be tracked in ticket #76.

Yes, you are right, I will link the PRs in ticket #76, then close this.

PR #92, opened 2026-05-22
Format-specific DCQL (supported VCs)

### LINKED ISSUE #72: Format-specific DCQL (supported VCs)
Description: Use format id dc+sd-jwt in DCQL and metadata (not legacy draft-era format strings). For jwt_vc_json credential queries, set required meta.type_values. For dc+sd-jwt credential queries, set required meta.vct_values. Use VC-root Claims Path Pointers for W3C VC (not paths that start from the VP wrapper).

Spec: Final 1.0 Appendix B.1 https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-w3c-verifiable-credentials / B.3 (as applicable) https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-ietf-sd-jwt-vc

Approach: Extend the DCQL builder and validation pipeline accordingly.

mbunwe-victor:
## Summary
Refactors the OpenID4VP DCQL flow to be format-capability-driven while keeping the production implementation focused on SD-JWT (`dc+sd-jwt`) for this PR.

This PR now intentionally narrows scope to the currently supported authentication pipeline and delivers the extensibility hooks needed for future formats.

## What changed
- Introduced a capability registry for format-specific DCQL behavior:
  - `DcqlCredentialCapability`
  - `DcqlCredentialCapabilities`
  - `SdJwtDcqlCredentialCapability` (current production capability)
- Kept DCQL query construction/validation focused on SD-JWT:
  - `SdJwtCredentialConstrainer`
  - `DcqlQueryValidator` (SD-JWT `meta.vct_values` rules)
  - `DcqlPresentationValidator` (SD-JWT presentation checks)
  - `ClaimPathResolver` (VC-root claim path resolution for SD-JWT)
- Completed response-side decoupling:
  - `AuthorizationResponseService` no longer hardcodes SD-JWT auth-note setup.
  - Authentication handoff is now owned by the resolved capability via `setupAuthenticationSession(...)`.
- Wired request, response, and discovery flows through the shared capability registry from `OID4VPUserAuthEndpoint`.

## Scope decisions for this PR
- Removed `jwt_vc_json` implementation code and tests introduced in this branch.
- Retains only the SD-JWT capability in `createDefault()`.
- This keeps the PR aligned with the current backend authentication support while preserving clean extension points for later formats.

## Follow-up
- `jwt_vc_json` (and other formats such as mso_mdoc) can be added in isolated follow-up PRs by implementing new `DcqlCredentialCapability` variants without changing core request/response services.

## Test plan
- [x] `DcqlQueryValidatorTest`
- [x] `DcqlCredentialCapabilitiesTest`
- [x] `SdJwtCredentialConstrainerTest`
- [x] Full `mvn test` in CI/local environment

Related: #72


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/DcqlPresentationValidator.java
```diff
@@ -0,0 +1,176 @@
...
+    private static JsonNode parseJwtPayload(String jwt) throws VerificationException {
+        try {
+            return new JWSInput(jwt).readJsonContent(JsonNode.class);
+        } catch (JWSInputException e) {
+            throw new VerificationException("Could not parse presented JWT payload", e);
+        }
+    }
+
+    private static JsonNode extractVcPayload(JsonNode vpPayload) throws VerificationException {
+        JsonNode verifiableCredentials = vpPayload.get(VP_VERIFIABLE_CREDENTIAL_CLAIM);
```
forkimenjeckayang: The `jwt_vc_json` validator appears to validate a flattened test-only JWT shape instead of the final 1.0 JWT VP / VC envelope. The spec examples place credentials under `vp.verifiableCredential`, and JWT VC content under the `vc` claim. Please update extraction so DCQL metadata and claim paths are applied to the W3C VC root, not the JWT payload root, and add tests using spec-shaped JWT VP and JWT VC payloads.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/VerifierDiscoveryService.java
```diff
@@ -43,9 +44,9 @@ public ClientMetadata getClientMetadata(EphemeralKey ephemeralKey) {
         logger.debug("Discovering Keycloak's metadata as an OpenID4VP client");
         ClientMetadata metadata = new ClientMetadata();
 
-        // Only SD-JWT presentations are supported for now.
         ClientMetadata.VpFormat vpFormat = new ClientMetadata.VpFormat();
         vpFormat.setDcSdJwt(getSdJwtVpFormat());
+        vpFormat.setJwtVcJson(getJwtVcJsonVpFormat());
```
forkimenjeckayang: Please do not advertise `jwt_vc_json` in verifier metadata until the production response path can authenticate it end-to-end. The current flow still routes every accepted token through `SdJwtAuthenticator`, and the JWT VC helper only parses payloads. Metadata should reflect actual verifier-supported formats, not future planned support.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/DcqlPresentationValidator.java
```diff
@@ -0,0 +1,176 @@
...
+        JsonNode typeNode = vcPayload.get("type");
+        if (typeNode == null || !typeNode.isArray() || typeNode.isEmpty()) {
+            throw new VerificationException("Presented jwt_vc_json credential is missing required type claim");
+        }
+
+        Set<String> presentedTypes = new HashSet<>();
+        typeNode.forEach(node -> presentedTypes.add(node.asText()));
+
+        boolean matches =
+                meta.getTypeValues().stream().anyMatch(requiredTypes -> presentedTypes.containsAll(requiredTypes));
```
forkimenjeckayang: The `jwt_vc_json` type matcher compares raw `type` strings, but final 1.0 evaluates W3C VC `type_values` after `@context` expansion. Please either perform JSON-LD expansion or implement an equivalent deterministic expansion for the supported contexts before applying `containsAll`.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/ClaimPathResolver.java
```diff
@@ -0,0 +1,66 @@
...
+        if (path.size() == 1) {
+            return hasDisclosedClaim(sdJwt, path.getFirst());
+        }
+
+        return false;
+    }
+
+    private static JsonNode resolveInPayload(JsonNode root, List<String> path) {
+        JsonNode current = root;
+        for (String segment : path) {
```
forkimenjeckayang, line 44: The DCQL path model and resolver only support string object keys. Final 1.0 Claims Path Pointers for JSON credentials also allow `null` array wildcards and non-negative integer indexes. Please either model full path components now or clearly scope this builder/validator to string-only paths and avoid presenting it as final 1.0 path support.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/DcqlQueryValidator.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private DcqlQueryValidator() {}
+
+    public static void validateQuery(DcqlQuery query) {
+        if (query == null
+                || query.getCredentials() == null
+                || query.getCredentials().isEmpty()) {
+            throw new IllegalArgumentException("dcql_query.credentials must be non-empty");
+        }
+        query.getCredentials().forEach(DcqlQueryValidator::validateCredential);
```
forkimenjeckayang, line 29: `DcqlQueryValidator` is missing several Section 6 structural checks, including credential id syntax/uniqueness and claim-set/id consistency. Since this validator is now in the signed request build path and is intended as shared DCQL validation, please enforce these constraints before accepting the query.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationRequestService.java
```diff
@@ -66,15 +67,15 @@ public class AuthorizationRequestService {
     public static final String SYMBOLIC_AUD = "https://self-issued.me/v2";
 
     private final KeycloakSession session;
-    private final SdJwtCredentialConstrainer constrainer;
+    private final SdJwtCredentialConstrainer sdJwtCredentialConstrainer;
```
forkimenjeckayang: This new DCQL path is still wired directly to SD-JWT request requirements. If this PR is meant to establish the format-specific DCQL foundation for JWT VC and future mdoc support, please introduce a small format capability/registry boundary instead of injecting `SdJwtCredentialConstrainer` directly here. Each capability should own its metadata, DCQL credential construction, and response/authentication handoff so adding `jwt_vc_json` or `mso_mdoc` does not require editing the core request service.

forkimenjeckayang (changes requested):
A few comments to check @mbunwe-victor

forkimenjeckayang (commented):
**Great work on the architectural direction of this PR!**
The introduction of the `DcqlCredentialCapability` registry is a good step forward. It cleanly abstracts the format-specific metadata advertisement and DCQL query construction, perfectly setting the stage for future formats.

However, after another look into the changes, there is a significant issue with **scope creep** regarding the `jwt_vc_json` implementation that we need to address before merging:

### 1. Remove all `jwt_vc_json` specific implementation code
FOr me, the goal of this architectural shift was to make the core services *extensible* so we can easily add new formats later—not to actually implement them right now. Currently, the plugin's authentication pipeline only supports SD-JWT.

By including `JwtVcJsonCredentialConstrainer`, `VcJsonLdTypeExpander`, and hardcoding `VCFormat.JWT_VC` switch branches inside `DcqlPresentationValidator` and `DcqlQueryValidator`, we are introducing a massive amount of dead code and JSON-LD maintenance burden for a format the backend cannot even authenticate yet.

Please completely remove all `jwt_vc_json` specific code, classes, and tests from this PR. The only capability implemented in this PR should be the `SdJwtDcqlCredentialCapability`. The beauty of the new SPI architecture is that when we *do* want to support `mso_mdoc` in the future, we can add them in isolated PRs without touching the core validators though with KC we support issunace of `jwt_vc_json` redemtials but I am not sure we want to include that now..

### 2. Complete the Decoupling (The Authenticator Handoff)
While the *request* side is now  decoupled via the capability registry, the *response* side in `AuthorizationResponseService.java` still explicitly hardcodes the handoff to the SD-JWT pipeline:

```java
// Still hardcoded to SdJwtAuthenticator
processorSession.setAuthNote(SdJwtAuthenticator.SDJWT_TOKEN_KEY, sdJwtVp);
```

If we leave this hardcoded, adding a new format capability later will still require editing this core service (otherwise it will crash trying to feed a different format into the SD-JWT authenticator).

Please extend the `DcqlCredentialCapability` interface to also own the authentication handoff. For example, add a method like `void setupAuthenticationSession(AuthenticationSessionModel session, String vpToken, AuthorizationContext context);`. `SdJwtDcqlCredentialCapability` can then set its own specific auth notes, entirely removing the hardcoded `SdJwtAuthenticator` logic from the core service.

Will like to take input from the others before proceeding with this
@IngridPuppet  @Awambeng  @Vitalisn4

Vitalisn4 (commented):
Hello @mbunwe-victor , please find additional comments for you to look at if they are okay to address with regards to changes

### PR description vs what the request object actually advertises

The PR summary says verifier metadata advertises both `dc+sd-jwt` and `jwt_vc_json`. With the default registry (`DcqlCredentialCapabilities.createDefault()`, which only registers `SdJwtDcqlCredentialCapability`), the happy path test shows something different on the signed request object:

- `dc+sd-jwt` is there in `vp_formats_supported`
- `jwt_vc_json` is not (`getJwtVcJson()` is null)

See `OID4VPUserAuthEndpointTest.shouldAuthenticateSuccessfully_NewDcSdJwtFormat`.

`ClientMetadataSerializationTest` is useful, but it only checks Jackson property names when both formats are set manually on the model. It does not go through `VerifierDiscoveryService.getClientMetadata()` with the default capabilities.

Could you update the PR description to match what we actually emit at runtime, and add a test that serializes metadata from `VerifierDiscoveryService` with the default registry? That would keep advertised formats tied to what is really implemented.

### Closing issue #72 if this stays SD-JWT only

Issue #72 also talks about `jwt_vc_json` and `meta.type_values` (Appendix B.1), not only `dc+sd-jwt` and `meta.vct_values` (Appendix B.3).

If this PR ends up SD-JWT only, as discussed in the earlier review, it would help to be explicit about how we close #72. Either narrow the acceptance criteria or open a follow-up for `jwt_vc_json`, so we do not imply the whole issue is done when only the SD-JWT path landed.

### Keep DCQL and the authenticator in sync

Login `dcql_query` is built from `SdJwtAuthRequirements.getSdJwtQuerySpec()` (vcts, `sub`, `username`, holder binding). `SdJwtAuthenticator` still does more on top of that (signatures, issuer, KB-JWT nonce/aud, revocation, and so on).

It might be worth a short comment or a small test that `getSdJwtQuerySpec()` stays the single place that decides which vcts and claim paths end up in DCQL. That should reduce the chance that a presentation passes DCQL validation but then fails later in the authenticator because the two layers drifted apart.

### Claims path support is intentionally limited

`DcqlQueryValidator` and `ClaimPathResolver` already document that we only support object property name segments right now (no array indexes, no `null` wildcards). In production we still use single segment paths like `sub` and `username` via `QuerySpec.of(...)`.

That is fine for the current login flow, but it is not the full Appendix B.1 story (for example nested paths like `credentialSubject.given_name`). A note in the PR or a follow-up issue would make it clear this is a known limit, not something we forgot.

### Why `credential_sets` is on a single credential query

`DcqlQueryBuilder.singleCredentialQuery` always adds `credential_sets` with one option. The HAIP tests depend on that shape, and `AuthorizationRequestService` already calls out that multi-credential would need more work on the response side.

A one line comment on `DcqlQueryBuilder` or in the PR description explaining why we emit `credential_sets` for a single credential (HAIP / wallet interoperability) would help the next person reading this.

### Missing test for holder binding at the DCQL layer

`require_cryptographic_holder_binding` defaults to true, and `DcqlPresentationValidator` can reject a presentation that has no KB-JWT.

I did not see an integration test that sends a presentation without KB-JWT and expects `invalid_vp_token` before `SdJwtAuthenticator` runs. Worth adding when you are in this area, especially if holder binding becomes configurable later.

Vitalisn4 (dismissed):
Thanks @mbunwe-victor for addressing the comments. Looks good from my side unless @forkimenjeckayang might still have some comments to add.

**Minor (non-blocking):**  Claims Path Pointer limits (no array indexes / null wildcards; single-segment paths) are still only in code javadoc. A short note in the PR or a follow-up issue would make that explicit for Appendix B.1

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/DcqlQueryBuilder.java
```diff
@@ -0,0 +1,38 @@
...
+    public static List<Claim> claimsFromPaths(List<List<String>> paths) {
+        return paths.stream().map(DcqlQueryBuilder::claimFromPath).toList();
+    }
+
+    public static Claim claimFromPath(List<String> path) {
+        Claim claim = new Claim();
+        claim.setId(UUID.randomUUID().toString());
+        claim.setPath(path);
+        return claim;
+    }
```
Awambeng, line 26: These utility methods aren’t being used. Do you think it’s worth keeping them?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/SdJwtDcqlCredentialCapability.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    public String format() {
+        return VCFormat.SD_JWT_VC;
+    }
+
+    @Override
+    public boolean supports(VerifierConfig config) {
+        return true;
+    }
```
Awambeng, line 30: This method always returns true. This works for now with only SD-JWT supported, but should become conditional once additional formats are introduced. Consider adding a // TODO for future implementors.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/DcqlPresentationValidator.java
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Validates that a presented credential satisfies the DCQL constraints from the authorization request.
+ */
+public final class DcqlPresentationValidator {
+
+    private DcqlPresentationValidator() {}
+
+    public static void validatePresentation(DcqlQuery query, String presentedToken) throws VerificationException {
+        DcqlQueryValidator.validateQuery(query);
```
Awambeng, line 23: This method already calls `DcqlQueryValidator.validateQuery(query)` before delegating to `validateSdJwtPresentation()`, which calls the same validation again, resulting in duplicate validation on the hot path. Consider removing the redundant call in `validateSdJwtPresentation()` since it should always be invoked through `validatePresentation()`.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtCredentialConstrainer.java
```diff
@@ -16,38 +10,16 @@
  */
 public class SdJwtCredentialConstrainer {
 
+    private final io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.dcql.SdJwtCredentialConstrainer delegate =
+            new io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.dcql.SdJwtCredentialConstrainer();
```
Awambeng: Please avoid using FQNand import the class correctly.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthRequirements.java
```diff
@@ -112,6 +113,14 @@ public PresentationRequirements getPresentationRequirements() {
...
 
+    public QuerySpec getSdJwtQuerySpec() {
+        return getSdJwtQuerySpec(requireCryptographicHolderBinding);
+    }
+
+    public QuerySpec getSdJwtQuerySpec(boolean requireHolderBinding) {
+        return QuerySpec.of(getExpectedVcts(), getRequiredClaims(), requireHolderBinding);
+    }
+
     public SdJwtCredentialConstrainer.QueryMap getSdJwtQueryMap() {
```
Awambeng: We probably don’t need QueryMap anymore in production code. Since it’s only used in tests, it may be worth simplifying the codebase by removing it

Awambeng (changes requested):
Hi @mbunwe-victor, thank you for the PR! 
I left a couple of comments could you please take a look when you get a chance?

mbunwe-victor:
Please could you check this again, @Awambeng?

Awambeng (dismissed):
Thanks for addressing my comments. LGTM.

Vitalisn4 (dismissed):
LGTM

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/DcqlPresentationValidator.java
```diff
@@ -0,0 +1,80 @@
...
+
+    private static void validateHolderBinding(Credential credentialQuery, SdJwtVP presentation)
+            throws VerificationException {
+        Boolean required = credentialQuery.getRequireCryptographicHolderBinding();
+        if (Boolean.TRUE.equals(required) && presentation.getKeyBindingJWT().isEmpty()) {
+            throw new VerificationException("DCQL query requires cryptographic holder binding (Key Binding JWT)");
+        }
+    }
+
+    private static void validateSdJwtMeta(Meta meta, SdJwtVP presentation) throws VerificationException {
```
forkimenjeckayang, line 52: According to the OpenID4VP 1.0 specification (Section 6.1), the `require_cryptographic_holder_binding` property is optional and **defaults to `true`**:

> `- require_cryptographic_holder_binding (OPTIONAL):`
>   - `boolean, default true.`

Currently, checking `Boolean.TRUE.equals(required)` means that if the parameter is omitted in the query (deserialized as `null`), key binding validation is bypassed.

To align with the default behavior of the spec, we should enforce validation unless it is explicitly disabled (`Boolean.FALSE`).

```suggestion
    private static void validateHolderBinding(Credential credentialQuery, SdJwtVP presentation)
            throws VerificationException {
        Boolean required = credentialQuery.getRequireCryptographicHolderBinding();
        if (!Boolean.FALSE.equals(required) && presentation.getKeyBindingJWT().isEmpty()) {
            throw new VerificationException("DCQL query requires cryptographic holder binding (Key Binding JWT)");
        }
    }

forkimenjeckayang (changes requested):
One last comment @mbunwe-victor

Vitalisn4 (approved):
LGTM

Awambeng (approved):

forkimenjeckayang (approved):

PR #94, opened 2026-05-26
feat(oid4vp): DCQL model and verifier-side satisfaction

mbunwe-victor:
This PR adds DCQL model extensions and verifier-side validation so the plugin can independently check that returned presentations satisfy the issued DCQL query (OpenID4VP §6.4, §8.6, §14.9), per [vp_token validation](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-vp-token-validation).

Validation follows OpenID4VP §8.6 ordering in the authorization response flow:

1. **Structural extraction** — `AuthorizationResponseService` checks the `vp_token` map shape, decodes the presentation, and verifies it is parseable as SD-JWT. No DCQL satisfaction at this stage.
2. **Integrity and authenticity** — `SdJwtAuthenticator` verifies signatures, issuer trust, and cryptographic holder binding via `verifySdJwtPresentation`. When a DCQL query is present, presentation requirements are limited to integrity-only checks (`getIntegrityOnlyPresentationRequirements()`); `vct` and claim constraints are deferred to step 3.
3. **DCQL satisfaction** — `SdJwtAuthenticator.validateDcqlSatisfaction()` reads the persisted `dcql_query` auth note and delegates to `DcqlCredentialCapability.validatePresentation()` → `DcqlSatisfactionValidator`. DCQL failures return `400 INVALID_VP_TOKEN`.

#### DCQL model

- `Claim.path` / `Claim.values` → `List<Object>` for typed DCQL constraints and §7.1 mixed path pointers (string keys, array indexes, wildcards)
- `Credential`: `require_cryptographic_holder_binding` (+ `requiresCryptographicHolderBinding()` helper)
- **`Credential.multiple` intentionally omitted** — deferred to a follow-up PR per [PR #86 discussion](https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/86#discussion_r3280559340); login flow enforces exactly one presentation per credential query id

#### Verifier-side DCQL validation (`validation` package)

- `DcqlSatisfactionValidator` — validates `meta.vct_values`, `claims`, `claim_sets`, and typed `values`
- `DcqlQueryRules` — validates `credential_sets` and credential-id selection rules
- `ClaimsPathProcessor` — DCQL §7.1 path pointer resolution
- `SdJwtClaimReader` + `SdJwtDisclosedPayloadAssembler` — claim extraction from SD-JWT presentations
- `DcqlClaimValues` — strict JSON-type matching for DCQL `values`
- `PresentedCredential` — parsed presentation bound to its DCQL credential query
- `VpTokenPresentationDecoder` — Base64URL decode helper for wallet-encoded presentations
- `VpTokenValidationException` (+ `Phase.STRUCTURE` / `Phase.DCQL`)
- `VpTokenDcqlValidator` — tested building block for multi-presentation `vp_token` flows; the OID4VP login path uses `DcqlCredentialCapability.validatePresentation()` directly with a single credential

#### Login path wiring

- `SdJwtDcqlCredentialCapability.setupAuthenticationSession()` persists the issued `dcql_query` for post-crypto validation
- `SdJwtAuthenticator` runs DCQL satisfaction after cryptographic verification and maps mismatches to `invalid_vp_token`
- `AuthorizationResponseService` maps authenticator `invalid_vp_token` errors to `ProcessingError.INVALID_VP_TOKEN`
- Removed duplicate validators: `DcqlPresentationValidator`, `ClaimPathResolver`
- Integration tests updated in `OID4VPUserAuthEndpointTest` — DCQL failures (wrong `vct`, missing required claim) occur after cryptographic verification

### Test plan

- [ ] Unit tests: `DcqlSatisfactionValidatorTest`, `ClaimsPathProcessorTest`, `DcqlClaimValuesTest`, `SdJwtClaimReaderTest`, `VpTokenDcqlValidatorTest`, `VpTokenPresentationDecoderTest`
- [ ] Integration: `OID4VPUserAuthEndpointTest` — wrong vct, missing required claim, non-matching credential id fail with `INVALID_VP_TOKEN` after crypto verification
- [ ] Happy path: successful OID4VP login still works with Base64URL-encoded `vp_token`


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/validation/DcqlSatisfactionValidator.java
```diff
@@ -0,0 +1,161 @@
...
+ * <p>The wallet may have enforced DCQL already; the verifier MUST NOT rely on that.
+ */
+public class DcqlSatisfactionValidator {
+
+    public void validate(List<PresentedCredential> presentations, DcqlQuery dcqlQuery)
+            throws VpTokenValidationException {
+        Map<String, PresentedCredential> presentedById = new HashMap<>();
+        for (PresentedCredential presented : presentations) {
+            presentedById.put(presented.credentialQueryId(), presented);
+        }
```
forkimenjeckayang, line 30: This map collapses duplicate presentations for the same DCQL credential id, so the validator does not enforce the Section 8.1 rule that `multiple=false` or omitted means exactly one presentation. It also does not reject unknown extra ids when `credential_sets` is absent. Please validate the returned `vp_token` cardinality/key set before reducing to a map.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/validation/DcqlQueryRules.java
```diff
@@ -0,0 +1,113 @@
...
+            if (options == null) {
+                continue;
+            }
+            for (List<String> option : options) {
+                if (option == null || option.isEmpty()) {
+                    continue;
+                }
+                if (presentedById.keySet().containsAll(option)) {
+                    permitted.addAll(option);
+                }
```
forkimenjeckayang: This currently permits the union of all satisfied credential-set alternatives. For `options=[["pid"], ["reduced_pid"]]`, returning both ids is accepted, even though the request asked for one acceptable combination. Please select/validate one option per satisfied credential set and reject extra returned credential ids that are outside the selected combination plus any satisfied optional sets.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/validation/DcqlSatisfactionValidator.java
```diff
@@ -0,0 +1,161 @@
...
+        if (!VCFormat.SD_JWT_VC.equals(expectedFormat) && !"dc+sd-jwt".equals(expectedFormat)) {
+            throw new VpTokenValidationException(
+                    VpTokenValidationException.Phase.DCQL,
+                    "Unsupported credential format in DCQL query: " + expectedFormat);
+        }
+    }
+
+    private void validateMeta(SdJwtVP presentation, Meta meta) throws VpTokenValidationException {
+        if (meta == null || meta.getVctValues() == null || meta.getVctValues().isEmpty()) {
+            return;
```
forkimenjeckayang: For `dc+sd-jwt`, `meta.vct_values` is required and non-empty. Returning here accepts an SD-JWT credential query without checking any credential type. Please reject missing/empty `vct_values` for SD-JWT format queries rather than treating it as unconstrained metadata.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/validation/ClaimsPathProcessor.java
```diff
@@ -0,0 +1,127 @@
...
+     * <p>Today {@code Claim.path} is {@code List<String>} because our DCQL schemas only use object-key
+     * segments. OpenID4VP §7.1 also allows integers (array indices) and {@code null} (all array elements),
+     * e.g. {@code ["address", null, "street_address"]}. {@link #process} already supports those types.
+     *
+     * <p>If {@code Claim.path} is widened to deserialize mixed JSON array elements, update this mapper to
+     * translate numeric indices and {@code null} entries—not only {@link List#copyOf} of strings.
+     */
+    public static List<Object> toPathComponents(List<String> path) {
+        return List.copyOf(path);
+    }
```
forkimenjeckayang: The processor supports mixed path components, but this adapter still takes `List<String>` from the DCQL model and only copies string segments. That means final-spec paths containing `null` or array indexes cannot reach the processor. Please widen the model/adapter contract and validate supported component types so the §7.1 processor is actually reachable for spec-valid paths.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/validation/ClaimsPathProcessor.java
```diff
@@ -0,0 +1,127 @@
...
+        throw new VpTokenValidationException(
+                VpTokenValidationException.Phase.DCQL, "Unsupported claims path pointer component: " + component);
+    }
+
+    private static List<JsonNode> selectObjectKey(List<JsonNode> current, String key) {
+        List<JsonNode> next = new ArrayList<>();
+        for (JsonNode node : current) {
+            if (!node.isObject()) {
+                continue;
+            }
```
forkimenjeckayang, line 65: For string path components, the spec says every current selected element must be an object; non-objects are an error. This branch silently skips non-object elements, so mixed arrays can still satisfy a path by ignoring invalid elements. Please make object-key selection strict, matching the array/index branches.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/validation/DcqlClaimValues.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+        if (expected instanceof Integer expectedInteger) {
+            return actual.isIntegralNumber() && actual.asInt() == expectedInteger;
+        }
+        if (expected instanceof Long expectedLong) {
+            return actual.isIntegralNumber() && actual.asLong() == expectedLong;
+        }
+        if (expected instanceof Number expectedNumber) {
+            return actual.isNumber() && actual.asDouble() == expectedNumber.doubleValue();
+        }
```
forkimenjeckayang: DCQL `values` only allows string, integer, and boolean entries. This branch accepts arbitrary numbers and compares them as doubles, so decimal constraints become valid even though they are outside the spec. Please reject non-integral numbers during query/value validation and keep matching integer-exact.

forkimenjeckayang (changes requested):
I left some comments @mbunwe-victor 
Please check

mbunwe-victor:
Please could you check this again, @forkimenjeckayang?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/validation/SdJwtClaimReader.java
```diff
@@ -0,0 +1,50 @@
...
+    public static ObjectNode disclosedClaimsRoot(SdJwtVP sdJwt) throws VerificationException {
+        Map<String, JsonNode> claims = new LinkedHashMap<>();
+        for (String disclosure : sdJwt.getDisclosuresString()) {
+            ArrayNode arrayNode = SdJwtUtils.decodeDisclosureString(disclosure);
+            if (arrayNode.size() == 3) {
+                claims.put(arrayNode.get(1).asText(), arrayNode.get(2));
+            }
+        }
+        ObjectNode root = com.fasterxml.jackson.databind.node.JsonNodeFactory.instance.objectNode();
+        claims.forEach(root::set);
```
forkimenjeckayang: This method extracts all SD-JWT disclosures and places them into a flat `Map` at the root level using `arrayNode.get(1).asText()` as the key. This approach causes two critical spec violations:
1. **Array Element Disclosures:** Array element disclosures (which have a size of 2: `[salt, claim_value]`) are completely ignored because of the `if (arrayNode.size() == 3)` check.
2. **Nested Disclosures:** Nested object disclosures are incorrectly flattened. If the issuer payload contains a nested digest (e.g., inside `address.street`), its disclosure will be mapped to the *root* of the assembled JSON rather than re-inserted into the `address` object. 

As a result, DCQL path queries targeting array elements or nested properties will fail to resolve because the payload hierarchy is not correctly reconstructed. You must rebuild the `ObjectNode` recursively according to the SD-JWT specification before applying the `ClaimsPathProcessor`.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/validation/DcqlSatisfactionValidator.java
```diff
@@ -0,0 +1,166 @@
...
+                    VpTokenValidationException.Phase.DCQL,
+                    "Presentation vct does not match any value requested in DCQL meta.vct_values");
+        }
+    }
+
+    private void validateClaims(SdJwtVP presentation, Credential credentialQuery) throws VpTokenValidationException {
+        List<Claim> claims = credentialQuery.getClaims();
+        if (claims == null || claims.isEmpty()) {
+            return;
+        }
```
forkimenjeckayang, line 99: By returning early here, the validator silently accepts an invalid DCQL query if it contains `claim_sets` but no `claims`. 
OpenID4VP Section 6.4.1 (Rule 4) explicitly states: **"`claim_sets` MUST NOT be present if `claims` is absent."**

Please add a check to actively reject the query (throw a `VpTokenValidationException`) if `credentialQuery.getClaimSets()` is non-empty when `claims` is absent, to enforce strict DCQL structure validation.

forkimenjeckayang (changes requested):
few comments @mbunwe-victor

forkimenjeckayang (dismissed):
LGTM, Unless @IngridPuppet has time for a final review. If not @Ogenbertrand  or @Awambeng please check. I'll wait for this to be merged so that i have a fixed number of files to focus on and review here https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/95

Ogenbertrand:
I fear this PR is doing morethan enough already i'll recommend you read the spec again: https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-vp-token-validation, sorry the discription of the ticket was misleading, this is what this ticket should validate in the plugin:

> 1. Validate the integrity and authenticity of the Presentation and Credential.
> 2. Validate that the returned Credential(s) meet all criteria defined in the query in the Authorization Request (e.g., Claims included in the presentation).

The PR satisfies this validation, but it authorization response flow never calls `DcqlSatisfactionValidator`.

@mbunwe-victor

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/dcql/Credential.java
Ogenbertrand, line 1: See the comments in this PR: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/86#discussion_r3280559340

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/dcql/Claim.java
Ogenbertrand, line 1: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/94/changes#r3361700505

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtCredentialConstrainer.java
Ogenbertrand, line 1: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/94/changes#r3361700505

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/dcql/Credential.java
mbunwe-victor: Following the PR #86 consensus with @IngridPuppet: **`Credential.multiple` has been removed** from the model. The login flow still enforces exactly one presentation per credential query id (`DcqlQueryRules.validatePresentationCounts` + `AuthorizationResponseService`). Multi-presentation support (`multiple: true`) is deferred to a follow-up PR once we agree on how to process multiple SD-JWT presentations in authentication.

DCQL satisfaction is now wired end-to-end: `AuthorizationResponseService` calls `VpTokenDcqlValidator`, which delegates to `DcqlSatisfactionValidator` before the SD-JWT authenticator runs.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/dcql/Claim.java
mbunwe-victor: `Claim.path` and `Claim.values` are widened to `List<Object>` because DCQL §7.1 path pointers are not strings-only—they include string keys, `null` (all array elements), and integer indexes. This is consumed by `ClaimsPathProcessor` during DCQL satisfaction checks (`DcqlSatisfactionValidator` → `SdJwtClaimReader`). Unlike `multiple`, this is actively used by the validator, not dead model surface.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtCredentialConstrainer.java
mbunwe-victor: `List.<Object>of(claimName)` matches the widened `Claim.path` type. The constrainer only emits simple string-key paths for the login flow today; spec-valid mixed paths (indexes, array wildcards) are handled when parsing external DCQL queries into the same model and validated via `ClaimsPathProcessor`.

mbunwe-victor:
> I fear this PR is doing morethan enough already i'll recommend you read the spec again: https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-vp-token-validation, sorry the discription of the ticket was misleading, this is what this ticket should validate in the plugin:
> 
> > 1. Validate the integrity and authenticity of the Presentation and Credential.
> > 2. Validate that the returned Credential(s) meet all criteria defined in the query in the Authorization Request (e.g., Claims included in the presentation).
> 
> The PR satisfies this validation, but it authorization response flow never calls `DcqlSatisfactionValidator`.
> 
> @mbunwe-victor

Hello @Ogenbertrand, thank you for this clarity, and this being the case, then it means once we close this PR, we can just close #95 and #96, right, since everything is handled as per the comment you shared. Could you check again?

Ogenbertrand:
I wouldn’t approve this PR yet. The idea is good because it improves DCQL validation, but the implementation has problems: it checks too early, repeats existing logic, and adds extra code that isn’t actually usable in the real authentication flow yet. Please carefully double check the PR again. You can meet @forkimenjeckayang or @IngridPuppet  for more clarity of what this PR should solve.

mbunwe-victor:
Checking this again after some discussion with @Ogenbertrand

PR #50, opened 2026-04-14
Bind OID4VP auth completion to the originating flow

Ogenbertrand:
This PR enforces single ownership of the OID4VP authentication flow.
- For browser/OIDC flows, authentication completion remains bound to the originating OIDC authentication session.
- For API-only flows, the `authorization_code` is no longer exposed through the public status endpoint and must instead be redeemed through a verifier-bound `/code` endpoint.

closes: https://github.com/ADORSYS-GIS/digital-identity-arch-review-mab/issues/11

PR #100, opened 2026-05-29
Protect OIDC-wrapped OpenID4VP flow with PKCE (#93)

### LINKED ISSUE #93: Protect OIDC-Wrapped OpenID4VP Authorization Code Flow with PKCE
Implement PKCE protection for the OIDC-wrapped OpenID4VP authentication flow to prevent unauthorized redemption of intercepted authorization codes.

Vitalisn4:
### Summary

Protects the **OIDC-wrapped OpenID4VP** login path with **OpenID4VP-owned PKCE** (#93)

The plugin now generates its own PKCE parameters when the wrapped login UI starts. The browser polls status, redeems the authorization code via `POST /oid4vp-auth/code` with the stored `code_verifier`, then submits the code to `oid4vp-auth-login`. This works **even when the parent OIDC authorize request had no PKCE**.

### Flow split

| Flow | How login completes | PKCE |
|------|---------------------|------|
| **OIDC-wrapped (browser / QR)** | Status poll → `POST /oid4vp-auth/code` → hidden form → `oid4vp-auth-login` | Generated in `OID4VPUserAuthBean`; verifier held in page JS |
| **OIDC-wrapped (same-device)** | Wallet redirect → callback URI → `oid4vp-auth-login` | Separate PKCE context created server-side at subflow start |
| **API-initiated** | `POST /oid4vp-auth/request` (PKCE required) → VP response → `POST /oid4vp-auth/code` | Unchanged; client supplies PKCE at `/request` |

### Security notes

- PKCE is **not** read from the parent OIDC session or authorize URL query when starting OpenID4VP login.
- The status endpoint **never** exposes `authorization_code`; redemption requires a valid `code_verifier`.
- `/oid4vp-auth/code` is **enabled** for wrapped OIDC contexts (browser polling depends on it).
- PKCE challenge is stored on the **authorization context**, not copied onto issued-code client session notes.
- `oid4vp-auth-login` still enforces parent-session binding via `PARENT_AUTH_SESSION_ID`; PKCE is enforced at `/code` redemption instead of a login-bridge note comparison.

### Files changed

**Production**
- `OID4VPUserAuthBean.java` — generate S256 PKCE per subflow; pass challenge to `startAuthentication`; expose `transactionId`, `codeVerifier`, and `/code` URL to the UI
- `OID4VPUserAuthEndpoint.java` — require PKCE at `startAuthentication`; hide `authorization_code` from status; allow `/code` for wrapped flows
- `AuthorizationResponseService.java` — stop persisting PKCE on intermediate code client session notes
- `OID4VPLoginActionsService.java` — remove parent-session PKCE note comparison (no longer needed)
- `login-oid4vp.ftl` / `oid4vp.js` — poll status, redeem via `/code`, then submit to login action

**Tests**
- `OID4VPUserAuthBeanTest` — assert generated PKCE is passed to `startAuthentication` (never null)
- `OID4VPLoginActionsServiceTest` — wrapped flow redeems via `/code`; login still provisions without parent OIDC PKCE
- `OID4VPBaseKeycloakTest` / `OID4VPBaseUserAuthEndpointTest` — test helpers for page-scraped verifier and status URL encoding

## Test plan

- [x] `./mvnw spotless:check`
- [x] `OID4VPUserAuthBeanTest` (5/5)
- [x] `OID4VPLoginActionsServiceTest` (9/9)
- [x] `OID4VPUserAuthEndpointTest` (35/35)
- [ ] Full `./mvnw test` (CI)

Closes #93


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
```diff
@@ -488,7 +495,7 @@ public AuthorizationContext startAuthentication(
 
         // Call delegate service to create an authorization request
         AuthorizationContext authorizationContext = authorizationRequestService.createAuthorizationRequest(
-                config, authSession, oidcAuthSession, codeChallengeDetails);
+                config, authSession, oidcAuthSession, effectiveCodeChallengeDetails);
```
forkimenjeckayang: Passing `effectiveCodeChallengeDetails` here means wrapped OIDC flows can store the parent OIDC PKCE challenge in the `AuthorizationContext`. Before this PR, wrapped contexts did not carry PKCE details, so `/oid4vp-auth/code` rejected them as "not configured for this flow". Now a completed wrapped transaction with `parentAuthSessionId` and PKCE notes can pass that API redemption check and return the authorization code after only `transaction_id + code_verifier`.

That creates a second completion path for wrapped login which does not run the checks added in `OID4VPLoginActionsService` (`PARENT_AUTH_SESSION_ID` and PKCE continuity against the active parent auth session). In the API flow this returned code is exchanged directly at the token endpoint, so wrapped flows should not become eligible for this endpoint just because they inherited the parent OIDC PKCE challenge. This also contradicts the PR summary that `/oid4vp-auth/request + /oid4vp-auth/code` is unchanged.

Please reject contexts with `parentAuthSessionId` in `redeemAuthorizationCode`, or otherwise keep API-initiated PKCE separate from the parent OIDC PKCE used only for continuity.

forkimenjeckayang (changes requested):

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/freemarker/OID4VPUserAuthBean.java
```diff
@@ -136,9 +137,14 @@ public AuthContextBean getAuthContext() {
     }
 
     private AuthorizationContext startOpenID4VPAuthentication(String clientId, boolean enableSameDeviceResponse) {
-        // TODO: Generate and pass code challenge details for ownership binding and enhanced security in started subflow
+        var params = session.getContext().getUri().getQueryParameters();
+        CodeChallengeDetails codeChallengeDetails = new CodeChallengeDetails(
+                params.getFirst(OAuth2Constants.CODE_CHALLENGE),
+                params.getFirst(OAuth2Constants.CODE_CHALLENGE_METHOD));
```
forkimenjeckayang: For wrapped OIDC flows, reading the `code_challenge` from the URI query parameters is conceptually incorrect and poses an injection risk. In standard Keycloak flows, the `code_challenge` is not forwarded to the `/login-actions/authenticate` URI; it is safely stored in the `AuthenticationSessionModel` client notes. 

By reading from the URI here, an attacker could append `&code_challenge=ATTACKER` to the login page URL, which would override the parent session's genuine PKCE parameters during the OpenID4VP subflow. 

Please pass `null` for `codeChallengeDetails` here and let the endpoint retrieve it exclusively from the parent `AuthenticationSessionModel`.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
```diff
@@ -593,6 +609,34 @@ private void validateOwnershipBinding(CodeChallengeDetails codeChallengeDetails)
         }
     }
 
+    private CodeChallengeDetails resolveCodeChallengeDetails(
+            OIDCAuthSession oidcAuthSession, CodeChallengeDetails codeChallengeDetails) {
+        if (codeChallengeDetails != null
+                && StringUtil.isNotBlank(codeChallengeDetails.codeChallenge())
+                && StringUtil.isNotBlank(codeChallengeDetails.codeChallengeMethod())) {
```
forkimenjeckayang: This prioritization logic allows the `codeChallengeDetails` (which currently can come from the URI query string in wrapped flows) to override the parent session's PKCE parameters. 

For wrapped flows (`oidcAuthSession != null`), the parent `AuthenticationSessionModel` should be the strict, single source of truth for PKCE. Consider restructuring this method so that if `oidcAuthSession != null`, it **always** reads from the parent session's client notes, ignoring any provided `codeChallengeDetails`. The provided `codeChallengeDetails` should only be used as the primary source for API-initiated flows.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/OID4VPLoginActionsService.java
```diff
@@ -148,6 +148,20 @@ public Response oid4vpAuthLogin(
...
+        String parentCodeChallengeMethod = authSession.getClientNote(OAuth2Constants.CODE_CHALLENGE_METHOD);
+        if (parentCodeChallenge != null || parentCodeChallengeMethod != null) {
+            String issuedCodeChallenge = result.getClientSession().getNote(OAuth2Constants.CODE_CHALLENGE);
+            String issuedCodeChallengeMethod = result.getClientSession().getNote(OAuth2Constants.CODE_CHALLENGE_METHOD);
+
+            if (!Objects.equals(parentCodeChallenge, issuedCodeChallenge)
+                    || !Objects.equals(parentCodeChallengeMethod, issuedCodeChallengeMethod)) {
+                return failOnInvalidCode(
+                        authSession, "Authorization code PKCE binding does not match this OIDC session");
+            }
```
forkimenjeckayang: By wrapping the continuity check in `if (parentCodeChallenge != null || ...)`, the validation is skipped entirely if the parent session does not use PKCE. This means if an intermediate code somehow acquires a PKCE challenge (e.g., through the URI injection mentioned earlier) when the parent session didn't request one, it will be accepted silently without validating continuity.

It's safer to perform a strict equality check regardless of whether the parent has PKCE or not, ensuring that the intermediate code exactly mirrors the parent session's PKCE state:
```java
String issuedCodeChallenge = result.getClientSession().getNote(OAuth2Constants.CODE_CHALLENGE);
String issuedCodeChallengeMethod = result.getClientSession().getNote(OAuth2Constants.CODE_CHALLENGE_METHOD);

if (!Objects.equals(parentCodeChallenge, issuedCodeChallenge)
        || !Objects.equals(parentCodeChallengeMethod, issuedCodeChallengeMethod)) {
    return failOnInvalidCode(authSession, "Authorization code PKCE binding does not match this OIDC session");
}
```

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/freemarker/OID4VPUserAuthBeanTest.java
```diff
@@ -124,6 +127,22 @@ public void shouldNotRecreateAuthContextInSameParsingSession() {
...
+
+        bean.getAuthContext();
+
+        ArgumentCaptor<CodeChallengeDetails> captor = ArgumentCaptor.forClass(CodeChallengeDetails.class);
+        verify(oid4vp, atLeastOnce())
+                .startAuthentication(eq(TEST_CLIENT_ID), nullable(OIDCAuthSession.class), captor.capture());
+
+        assertTrue(captor.getAllValues().stream().allMatch(details -> codeChallenge.equals(details.codeChallenge())));
+        assertTrue(captor.getAllValues().stream()
+                .allMatch(details -> OAuth2Constants.PKCE_METHOD_S256.equals(details.codeChallengeMethod())));
```
forkimenjeckayang: This test asserts that the bean reads PKCE parameters from the URI query. Since we should rely on the `AuthenticationSessionModel` instead of the URI to prevent PKCE spoofing in wrapped flows, this test should be removed or updated to verify that the bean passes `null` for `codeChallengeDetails` when initiating the subflow.

forkimenjeckayang (changes requested):
More comments @Vitalisn4

forkimenjeckayang (dismissed):
No further concerns unless the other reviewers have

mbunwe-victor (dismissed):
LGTM
But could you consider updating the description on the PR

Ogenbertrand (commented):
I have carefully analyze this code changes and this is my take for this PR:
The code still exposes the wrapped-flow auth_code from the status endpoint, and relies on `oid4vp-auth-login` to make that code useful only inside the original browser OIDC session. That is a resonable direction, but the PKCE binding check currently allows `null == null`:

```java
if (!Objects.equals(parentCodeChallenge, issuedCodeChallenge)
        || !Objects.equals(parentCodeChallengeMethod, issuedCodeChallengeMethod)) {
```
in the `OID4VPLoginActionsService` class.

So if the parent OIDC session has no PKCE notes, and the issued OID4VP code has no PKCE notes, the check passes. In that case, the new PKCE protection is not actually enforced. The protection falls back mainly to parent auth-session binding and Keycloak login-action/session-code checks. I would ask for a small hardening change here.

And ideally also reject missing PKCE earlier in `OID4VPUserAuthEndpoint.startAuthentication` for OIDC-wrapped flows, so invalid contexts are never created.

The PR is on the right track and removes the free /code redemption path, but it should require non-blank PKCE for wrapped OIDC flows and add a mismatch/missing-PKCE test before we say it fully resolves the security issue.

Please see this comments and what was already covered in this PR: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/50#pullrequestreview-4186884629, to know what was missing.

Ogenbertrand (dismissed):
The implementation LGTM.
Thank you @Vitalisn4.

Ogenbertrand:
we shouldn't merge this yet, this is some security issues, so i'll like @IngridPuppet to check before we merge.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -230,6 +231,12 @@ private String produceAuthorizationCode(
             clientSession.setNote(PARENT_AUTH_SESSION_ID, authContext.getParentAuthSessionId());
         }
 
+        if (StringUtil.isNotBlank(authContext.getCodeChallenge())
+                && StringUtil.isNotBlank(authContext.getCodeChallengeMethod())) {
+            clientSession.setNote(OAuth2Constants.CODE_CHALLENGE, authContext.getCodeChallenge());
+            clientSession.setNote(OAuth2Constants.CODE_CHALLENGE_METHOD, authContext.getCodeChallengeMethod());
+        }
```
IngridPuppet: Please check my other comment on a possible solution without relying on session notes. They are hard to track in general.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
```diff
@@ -445,6 +445,15 @@ public Response redeemAuthorizationCode(
...
         }
 
+        // This endpoint is intended for API-initiated flows only. Wrapped OIDC flows must complete through
+        // oid4vp-auth-login to preserve browser session binding and continuity checks.
+        if (StringUtil.isNotBlank(authorizationContext.getParentAuthSessionId())) {
+            throw new BadRequestException(errorResponse(
+                    Response.Status.BAD_REQUEST,
+                    OAuthErrorException.INVALID_REQUEST,
+                    "Authorization code redemption is not configured for this flow"));
+        }
```
IngridPuppet: This ticket precisely aims to make this endpoint accessible and used by the polling flow in the browser.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
IngridPuppet: This ticket should make it possible to no longer include the authorization code at this level, but only redeemable with PKCE.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
```diff
@@ -593,6 +601,26 @@ private void validateOwnershipBinding(CodeChallengeDetails codeChallengeDetails)
...
+            if (StringUtil.isBlank(codeChallenge) || StringUtil.isBlank(codeChallengeMethod)) {
+                throw new IllegalArgumentException(
+                        "OIDC session must include code_challenge and code_challenge_method (S256) for OpenID4VP login");
+            }
+
+            return new CodeChallengeDetails(codeChallenge, codeChallengeMethod);
+        }
+
+        return codeChallengeDetails;
+    }
```
IngridPuppet: This method makes it possible to get code challenge details from session notes. We should not rely on this mechanism. It is very easy to simply generate those details and pass them when calling `startAuthentication` in OIDC-initiated flows.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/freemarker/OID4VPUserAuthBean.java
```diff
@@ -136,7 +136,6 @@ public AuthContextBean getAuthContext() {
     }
 
     private AuthorizationContext startOpenID4VPAuthentication(String clientId, boolean enableSameDeviceResponse) {
-        // TODO: Generate and pass code challenge details for ownership binding and enhanced security in started subflow
         return oid4vp.startAuthentication(
                 clientId, new OIDCAuthSession(authSessionId, getLoginActionUrl(), enableSameDeviceResponse), null);
```
IngridPuppet: Exactly here. The last must not be null. Code challenge details can be generated in the module and the code verifier passed to the UI JavaScript for redeeming the authorization code.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/OID4VPLoginActionsService.java
```diff
@@ -168,6 +174,27 @@ public Response oid4vpAuthLogin(
...
+        if (StringUtil.isBlank(parentCodeChallenge) || StringUtil.isBlank(parentCodeChallengeMethod)) {
+            return false;
+        }
+        if (StringUtil.isBlank(issuedCodeChallenge) || StringUtil.isBlank(issuedCodeChallengeMethod)) {
+            return false;
+        }
+
+        return Objects.equals(parentCodeChallenge, issuedCodeChallenge)
+                && Objects.equals(parentCodeChallengeMethod, issuedCodeChallengeMethod);
+    }
```
IngridPuppet: That means if the parent OIDC session has no PKCE protection, then our OpenID4VP child session does not have any either. This is not wanted IMO, especially because it is very easy to always protect the child session irrespectively of the parent session. Eventually, I see no need to change anything in this class.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPBaseKeycloakTest.java
```diff
@@ -218,13 +218,39 @@ protected HttpResponse redeemAuthorizationCodeResponse(String transactionId, Str
     /**
      * Scrapes the action URL of the OpenID4VP login form.
      */
-    protected FormData getFreshOid4vpFormActionUrl() throws IOException {
-        String authEndpoint = Objects.requireNonNull(new URIBuilder(getAuthEndpointURI())
+    protected record WrappedOidcAuthorizeRequest(URI uri, String oidcPkceCodeVerifier) {}
```
IngridPuppet, line 229: The old comment does not match the new record.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPBaseKeycloakTest.java
```diff
@@ -218,13 +218,39 @@ protected HttpResponse redeemAuthorizationCodeResponse(String transactionId, Str
...
+        String codeVerifier = null;
+        if (includePkce) {
+            codeVerifier = PkceUtils.generateCodeVerifier();
+            String codeChallenge = PkceUtils.encodeCodeChallenge(codeVerifier, OAuth2Constants.PKCE_METHOD_S256);
+            builder.addParameter(OAuth2Constants.CODE_CHALLENGE, codeChallenge)
+                    .addParameter(OAuth2Constants.CODE_CHALLENGE_METHOD, OAuth2Constants.PKCE_METHOD_S256);
+        }
+
+        return new WrappedOidcAuthorizeRequest(builder.build(), codeVerifier);
+    }
```
IngridPuppet, line 250: It should not be needed to test this scenario where code challenge params are present in the triggerring OIDC URL. With the auth code no longer accessible via the status endpoint, it should be enough to just ensure that we can still complete the flow.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPLoginActionsServiceTest.java
```diff
@@ -135,6 +143,50 @@ public void shouldFailAuthentication_IfNonOid4vpCode() throws Exception {
                 authCode, "Authorization code was not issued upon OpenID4VP authentication");
     }
 
+    @Test
+    public void shouldNotProvisionOid4vpLogin_WhenOidcAuthorizeWithoutPkce() throws Exception {
```
IngridPuppet: But we still want Openid4VP login to proceed even if OIDC was triggered without PKCE. We protect our child Openid4VP login with self-generated PKCE parameters regardless.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPLoginActionsServiceTest.java
```diff
@@ -135,6 +143,50 @@ public void shouldFailAuthentication_IfNonOid4vpCode() throws Exception {
...
+            HttpResponse response = client.execute(httpGet);
+            int status = response.getStatusLine().getStatusCode();
+            if (status == HttpStatus.SC_OK) {
+                String html = EntityUtils.toString(response.getEntity());
+                assertNull(
+                        Jsoup.parse(html).selectFirst("form#kc-oid4vp-completion-form"),
+                        "OpenID4VP QR login must not start without PKCE on the parent OIDC authorize request");
+            } else {
+                assertTrue(
+                        status >= HttpStatus.SC_BAD_REQUEST,
```
IngridPuppet: In general, a test should not have if/else branches because it ends up unreliable. We finally don't know what we are testing. Should the status be `SC_OK` or not? The answer can't be both.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPLoginActionsServiceTest.java
```diff
@@ -135,6 +143,50 @@ public void shouldFailAuthentication_IfNonOid4vpCode() throws Exception {
...
+                        "Missing parent PKCE should fail before provisioning the OpenID4VP login UI");
+            }
+        }
+    }
+
+    @Test
+    public void shouldRejectApiCodeRedemption_ForWrappedOidcFlow() throws Exception {
+        FormData formData = getFreshOid4vpFormActionUrl();
```
IngridPuppet, line 169: We precisely want to enforce the contrary. OIDC-wrapped flows must be forced to redeem auth codes, not get them from the status endpoint.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/freemarker/OID4VPUserAuthBeanTest.java
```diff
@@ -124,6 +126,16 @@ public void shouldNotRecreateAuthContextInSameParsingSession() {
...
     }
 
+    @Test
+    public void shouldNotReadPkceParamsFromOidcQuery() {
+        OID4VPUserAuthBean bean = createTestBeanWithPkce("test-code-challenge", OAuth2Constants.PKCE_METHOD_S256);
+
+        bean.getAuthContext();
+
+        verify(oid4vp, atLeastOnce())
+                .startAuthentication(eq(TEST_CLIENT_ID), nullable(OIDCAuthSession.class), isNull());
```
IngridPuppet: PKCE details passed to `startAuthentication` must precisely never be null.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/OID4VPLoginActionsServicePkceBindingTest.java
IngridPuppet, line 1: Whenever possible, we should avoid testing with Mockito, especially on a path that can be covered by integration tests. It is very easy to end up testing nothing when relying on Mockito excessively.

IngridPuppet (changes requested):
I reviewed the changes and appreciate that, in general, changes in the PR are definitely meaningful. However, the ticket appears to have been very unclear. I've tried to add further clarifications in my comments. Please could you check?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/freemarker/OID4VPUserAuthBean.java
```diff
@@ -115,29 +117,49 @@ public AuthContextBean getAuthContext() {
...
-        // Initiate OID4VP authentication
+        // Initiate OID4VP authentication (cross-device QR flow drives status polling + /code redemption)
         String clientId = params.getFirst(OAuth2Constants.CLIENT_ID);
-        AuthorizationContext authContext = startOpenID4VPAuthentication(clientId, false);
-        AuthorizationContext authContextSameDevice = startOpenID4VPAuthentication(clientId, true);
+        OpenId4vpPkce crossDevicePkce = OpenId4vpPkce.generate();
+        AuthorizationContext authContext =
+                startOpenID4VPAuthentication(clientId, false, crossDevicePkce.challengeDetails());
+        AuthorizationContext authContextSameDevice = startOpenID4VPAuthentication(
+                clientId, true, OpenId4vpPkce.generate().challengeDetails());
```
IngridPuppet: Same-device flows don't need PKCE. They can access the auth code on redirection directly.

src/main/resources/theme/keycloak.v2+oid4vp/login/login-oid4vp.ftl
```diff
@@ -27,11 +27,19 @@
               method="post"
               style="display:none;">
             <input type="hidden" id="kc-oid4vp-code-input" name="code" value="" />
+            <input type="hidden" id="kc-oid4vp-transaction-id" value="${oid4vp.authContext.transactionId!""}" />
+            <input type="hidden" id="kc-oid4vp-code-verifier" value="${oid4vp.authContext.codeVerifier!""}" />
```
IngridPuppet: Could we avoid these hidden fields? They travel with the form response. You could declare variables in the script if the point is to easily scrape them in tests.

IngridPuppet (commented):
All great updates! Just minor more comments.

IngridPuppet (approved):
Thank you for the update.

PR #101, opened 2026-06-03
Add configurable OpenID4VP authentication profiles with dual credential support

### LINKED ISSUE #99: Extend the plugin to support requesting two VCs during the verification/presentation flow.
ETA: 2days

Ogenbertrand:
This PR adds support for multiple OpenID4VP authentication profiles in the plugin.

The main idea is to keep the existing single credential login untouched, while allowing a separate profile for dual credential login. So instead of changing the current wallet login behavior globally, the plugin now lets us configure different login profiles, for example:

Sign in with wallet: requests one credential
Dual login: requests two credentials
Each profile defines the credentials it needs, which one is the authenticating credential used to bind the Keycloak user, and which extra credentials are used only to strengthen the authentication.

This answers the main design question we had at the beginning: single and multiple VC flows are differentiated by profile selection. The user chooses the profile from the login page, and once that flow starts, the generated authorization request is tied to that selected profile. That prevents starting a dual credential flow and then downgrading to a single credential response.

The dual credential flow is fully configurable, so it is not limited to PID + Tax Advisor. A profile can define different credential IDs, formats, trust policies, claims, and binding rules.

In OpenID4VP authentication profiles, add a JSON config like:
```json
[
  {
    "id": "default",
    "displayCta": {
      "en": "Sign in with a wallet"
    },
    "credentials": [
      {
        "id": "identity",
        "role": "primary",
        "vct": ["<YOUR_EXISTING_VCT>"],
        "claims": ["sub", "username"]
      }
    ]
  },
  {
    "id": "dual",
    "displayCta": {
      "en": "Dual login"
    },
    "credentials": [
      {
        "id": "primary",
        "role": "primary",
        "vct": ["<MAIN_CREDENTIAL_VCT>"],
        "claims": ["sub", "username"]
      },
      {
        "id": "supporting",
        "role": "supporting",
        "vct": ["<SECOND_CREDENTIAL_VCT>"],
        "claims": ["username"],
        "trust": [
          { "type": "self" }
        ],
        "binding": [
          {
            "type": "claim_equals_primary_claim",
            "credentialClaim": "username",
            "primaryCredentialClaim": "username"
          }
        ]
      }
    ]
  }
]
```

image

Closes: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/99


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthenticator.java
```diff
@@ -136,6 +140,18 @@ public void action(AuthenticationFlowContext context) {
     }
 
     private SdJwtAuthRequirements getAuthenticationRequirements(AuthenticationFlowContext context) {
+        String credentialRequirement = context.getAuthenticationSession().getAuthNote(CREDENTIAL_REQUIREMENT_KEY);
+        if (StringUtil.isNotBlank(credentialRequirement)) {
```
IngridPuppet: `SdJwtAuthRequirements` should have direct access to this information without passing through a session note. Can't it? The flow with session notes is hard to track, so should only be used when no other readable option is viable.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthRequirements.java
```diff
@@ -83,14 +86,22 @@ public SdJwtAuthRequirements(KeycloakContext context, AuthenticatorConfigModel a
...
     public List<String> getExpectedVcts() {
         return expectedVcts;
     }
 
     public List<String> getRequiredClaims() {
-        // A subject is required so we can recover the user by stable identifier
-        // A username is required so we can cross-check the presented user
-        return List.of(JsonWebToken.SUBJECT, OAuth2Constants.USERNAME);
+        return requiredClaims;
     }
```
IngridPuppet, line 102: Do we check somewhere that the primary credential requests these two fields? Maybe not here, but we should have it enforced somewhere.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthRequirements.java
```diff
@@ -83,14 +86,22 @@ public SdJwtAuthRequirements(KeycloakContext context, AuthenticatorConfigModel a
...
 
+    public SdJwtAuthRequirements(
+            KeycloakContext context, AuthenticatorConfigModel authConfig, CredentialRequirement credentialRequirement) {
+        this(context, authConfig);
+        this.expectedVcts = credentialRequirement.getVct();
+        this.requiredClaims = credentialRequirement.getClaims();
+        this.expectedVctsPattern = expectedVcts.stream()
+                .map(vct -> Pattern.quote("\"" + vct + "\""))
+                .collect(Collectors.joining("|", "(", ")"));
```
IngridPuppet: This is repeated. Add a method and reuse?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtCredentialConstrainer.java
```diff
@@ -50,6 +52,57 @@ public DcqlQuery generateDcqlQuery(QueryMap queryMap) {
...
+        Credential credential = new Credential();
+        credential.setId(requirement.getId());
+        credential.setFormat(VCFormat.SD_JWT_VC);
+        credential.setMeta(meta);
+        credential.setClaims(claims);
+        if (!requireCryptographicHolderBinding) {
+            credential.setRequireCryptographicHolderBinding(false);
+        }
+        return credential;
+    }
```
IngridPuppet: This duplicates lots of code from `generateDcqlQuery`. Refactor?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/AuthenticationProfile.java
```diff
@@ -0,0 +1,87 @@
+package io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.profile;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+
+public class AuthenticationProfile {
+
```
IngridPuppet, line 10: Does it make sense to nest the entire `profile` package under `model`?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/CredentialRequirement.java
```diff
@@ -0,0 +1,92 @@
...
+    private String id;
+
+    @JsonProperty("role")
+    private CredentialRole role = CredentialRole.SUPPORTING;
+
+    @JsonProperty("formats")
+    private List<String> formats = List.of(VCFormat.SD_JWT_VC);
+
+    @JsonProperty("vct")
+    private List<String> vct;
```
IngridPuppet: The `vct` naming is SD-JWT-specific. mDoc will use `docType`. Could you use a format agnostic naming for this field?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/OID4VPProfileConfig.java
```diff
@@ -0,0 +1,121 @@
...
+ * Parses and validates OpenID4VP authentication profiles.
+ */
+public class OID4VPProfileConfig {
+
+    private final List<AuthenticationProfile> profiles;
+
+    public OID4VPProfileConfig(KeycloakContext context, AuthenticatorConfigModel authConfig) {
+        Map<String, String> config =
+                (authConfig != null && authConfig.getConfig() != null) ? authConfig.getConfig() : Map.of();
+        String configuredProfiles = config.get("profiles");
```
IngridPuppet: Reuse the config key constant.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -72,18 +91,33 @@ public void processAuthorizationResponse(
...
         String aud = authContext.getRequestObject().getClientId();
-        processorSession.setAuthNote(SdJwtAuthenticator.SDJWT_TOKEN_KEY, sdJwtVp);
+        processorSession.setAuthNote(SdJwtAuthenticator.SDJWT_TOKEN_KEY, primarySdJwtVp);
         processorSession.setAuthNote(SdJwtAuthenticator.CHALLENGE_NONCE_KEY, nonce);
         processorSession.setAuthNote(SdJwtAuthenticator.CHALLENGE_AUD_KEY, aud);
+        try {
+            processorSession.setAuthNote(
+                    SdJwtAuthenticator.CREDENTIAL_REQUIREMENT_KEY,
+                    JsonSerialization.writeValueAsString(primaryCredential));
+        } catch (Exception e) {
```
IngridPuppet: Use `valueAsString` to swallow the exception at this point. `primaryCredential` is already validated at this point. Plus, the user should not be blamed for serialization failures.
```suggestion
                    JsonSerialization.valueAsString(primaryCredential));
```

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -125,6 +159,16 @@ public void processAuthorizationResponse(
...
+                primaryCredential,
+                primarySdJwt,
+                clientSession.getUserSession().getUser(),
+                authConfig,
+                authContext,
+                store);
+
         // Produce an authorization code for the authenticated user
         String authorizationCode = produceAuthorizationCode(clientSession, authContext);
         authContext.setStatus(AuthorizationContextStatus.SUCCESS);
```
IngridPuppet, line 127: The logic is essentially to authenticate with the primary SD-JWT, then verify additional supporting credentials. I believe this breaks the architecture of the plugin and that of Keycloak regarding authentication. Technically, an authentication should never be marked successful prior to all required checks, including verifying supporting credentials. So the logic implemented should be internal to authenticators in the authenticator flow. Because we are only dealing with SD-JWTs for now. We could do it all in `SdJwtAuthenticator` and improve the code in another PR.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -152,71 +196,178 @@ private static String getAuthenticatorErrorMessage(Response response) {
...
+        var vpTokenMap = responseObject.getVpToken();
+        HashMap<String, String> tokens = new HashMap<>();
+
+        if (vpTokenMap == null) {
+            failInvalidVpToken("Presented vp_token map is missing", authContext, store);
+        }
+
+        for (CredentialRequirement credential : profile.getCredentials()) {
+            var credentialTokens = vpTokenMap.get(credential.getId());
+            if (credentialTokens == null || credentialTokens.size() != 1) {
```
IngridPuppet, line 167: Just to confirm, are you using the same IDs in the profile config in the DCQL query?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -152,71 +196,178 @@ private static String getAuthenticatorErrorMessage(Response response) {
...
+                authContext,
+                store);
+    }
+
+    private CredentialRequirement getPrimaryCredential(AuthenticationProfile profile) {
+        return profile.getCredentials().stream()
+                .filter(credential -> CredentialRole.PRIMARY.equals(credential.getRole()))
+                .findFirst()
+                .orElseThrow(() -> new IllegalStateException("Profile has no primary credential: " + profile.getId()));
+    }
```
IngridPuppet: Could this live directly in AuthenticationProfile?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -152,71 +196,178 @@ private static String getAuthenticatorErrorMessage(Response response) {
...
+            CredentialRequirement primaryCredential,
+            SdJwtVP primarySdJwt,
+            UserModel user,
+            AuthenticatorConfigModel authConfig,
+            AuthorizationContext authContext,
+            AuthenticationSessionStore store) {
+        for (CredentialRequirement credential : profile.getCredentials()) {
+            if (credential.getId().equals(primaryCredential.getId())) {
+                continue;
+            }
```
IngridPuppet: It looks like the method accepts `primaryCredential` just for this check. This can't be checked on CredentialRequirement directly?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -152,71 +196,178 @@ private static String getAuthenticatorErrorMessage(Response response) {
...
-                    "Invalid vp_token",
-                    detailed,
-                    Response.Status.BAD_REQUEST,
+                    ProcessingError.VP_TOKEN_AUTH_ERROR,
+                    "Invalid SD-JWT presentation",
+                    "Supporting credential '%s' could not be verified".formatted(credential.getId()),
+                    Response.Status.UNAUTHORIZED,
                     authContext,
                     store);
         }
```
IngridPuppet: With proper refactoring, we should aim for not duplicating this logic across the code base. I can see this largely overlaps with the logic in SdJwtAuthenticator.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -152,71 +196,178 @@ private static String getAuthenticatorErrorMessage(Response response) {
...
 
-        // Check that the VP token map provides a VP token, and only one
-        var tokens = vpTokenMap.get(credentialQuery.getId());
-        if (tokens == null || tokens.size() != 1) {
-            String errorMsg = String.format(
-                    "Presented vp_token map must contain exactly one token as requested. Found: %d",
-                    tokens == null ? 0 : tokens.size());
-
+    private void enforceSupportedTrustPolicy(
+            CredentialRequirement credential, AuthorizationContext authContext, AuthenticationSessionStore store) {
```
IngridPuppet: Update the naming to only prompt self trust?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -152,71 +196,178 @@ private static String getAuthenticatorErrorMessage(Response response) {
...
+                        case BindingRule.CLAIM_EQUALS_USER_ATTRIBUTE ->
+                            readUserAttribute(user, rule.getUserAttribute());
+                        default ->
+                            throw failWithHttpException(
+                                    ProcessingError.VP_TOKEN_AUTH_ERROR,
+                                    "Invalid SD-JWT presentation",
+                                    "Unsupported binding rule type: " + rule.getType(),
+                                    Response.Status.BAD_REQUEST,
+                                    authContext,
+                                    store);
```
IngridPuppet: The user should not be blamed for this. This is a case of IllegalStateException because the profile is considered validated at this point.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -152,71 +196,178 @@ private static String getAuthenticatorErrorMessage(Response response) {
...
 
-        return (String) tokens.getFirst();
+    private String readUserAttribute(UserModel user, String userAttribute) {
+        return switch (userAttribute) {
+            case "given_name", "firstName" -> user.getFirstName();
+            case "family_name", "lastName" -> user.getLastName();
+            case "username", "preferred_username" -> user.getUsername();
+            default -> user.getFirstAttribute(userAttribute);
+        };
     }
```
IngridPuppet: This could be done in a separate class to declutter this service.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpoint.java
```diff
@@ -485,14 +496,16 @@ public AuthorizationContext startAuthentication(
...
         AuthorizationContext authorizationContext = authorizationRequestService.createAuthorizationRequest(
-                config, authSession, oidcAuthSession, codeChallengeDetails);
+                config, profile, authSession, oidcAuthSession, codeChallengeDetails);
 
         return new AuthorizationContext()
                 .setAuthorizationRequest(authorizationContext.getAuthorizationRequest())
-                .setTransactionId(authorizationContext.getTransactionId());
+                .setTransactionId(authorizationContext.getTransactionId())
+                .setProfileId(authorizationContext.getProfileId());
     }
```
IngridPuppet: Any reason for leaking the profile ID?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/freemarker/OID4VPUserAuthBean.java
```diff
@@ -62,6 +64,16 @@ public OID4VPUserAuthBean(
      * URL to trigger UI view for signing in with a wallet
      */
     public String getLoginUrl() {
+        return getLoginProfiles().stream()
+                .findFirst()
+                .map(LoginProfileBean::getLoginUrl)
+                .orElse(null);
+    }
```
IngridPuppet: What's the point of keeping this?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -152,71 +196,178 @@ private static String getAuthenticatorErrorMessage(Response response) {
     /**
      * Extract SD-JWT VP token from response object
      */
-    private String extractSdJwtVpToken(
-            ResponseObject responseObject, AuthorizationContext authContext, AuthenticationSessionStore store) {
-        String parsedVpToken;
-        logger.debug("Extracting SD-JWT VP token from response object with DCQL matching");
-        parsedVpToken = extractSdJwtVpTokenWithDCQL(responseObject, authContext, store);
+    private HashMap<String, String> extractSdJwtVpTokens(
```
IngridPuppet, line 153: Update comment. Clarify the need for a new map beyond the one in response object.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/OID4VPProfileConfigTest.java
```diff
@@ -0,0 +1,70 @@
...
+                  {
+                    "id": "broken",
+                    "credentials": [
+                      { "id": "one", "role": "supporting", "vct": ["vct"], "claims": ["username"] }
+                    ]
+                  }
+                ]
+                """));
+
+        assertThrows(IllegalStateException.class, () -> new OID4VPProfileConfig(null, config));
```
IngridPuppet: Assert the exception message.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointTest.java
```diff
@@ -52,6 +53,8 @@
 public class OID4VPUserAuthEndpointTest extends OID4VPBaseUserAuthEndpointTest {
 
     public static final String VCT_CONFIG_ALT = "https://example.com/vct-alt";
+    public static final String DUAL_PROFILE_ID = "dual";
+    private static final String TEST_REALM_SD_JWT_AUTH_CONFIG_ID = "81d62be9-cf06-4718-8837-fdfb4727b20a";
```
IngridPuppet: Maybe also change the value to something more readable.

IngridPuppet (changes requested):
Great contribution! I left some comments, mainly regarding not breaking architectural agreements. Please could you check? 

Also, I don't recall seeing tests checking the creation of multiple buttons for signing in with respect to configured auth profiles.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthenticator.java
Ogenbertrand: fixed

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtCredentialConstrainer.java
Ogenbertrand: done

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthRequirements.java
Ogenbertrand: done

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/AuthenticationProfile.java
Ogenbertrand: The profile package isn't inside model :thinking:

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/CredentialRequirement.java
Ogenbertrand: yes done

Ogenbertrand:
Hi @IngridPuppet can you check if all your comments has been addressed?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthenticator.java
```diff
@@ -61,9 +66,9 @@ public class SdJwtAuthenticator implements Authenticator {
     public static final String SDJWT_TOKEN_KEY = "sdjwt_token";
 
     /**
-     * Optional serialized profile credential requirement to use instead of legacy flat config.
+     * Optional serialized map of DCQL credential IDs to presented SD-JWT VP tokens.
      */
-    public static final String CREDENTIAL_REQUIREMENT_KEY = "sdjwt_credential_requirement";
+    public static final String SDJWT_TOKENS_KEY = "sdjwt_tokens";
```
IngridPuppet, line 71: Any reason why these are sent separately?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/freemarker/OID4VPUserAuthBean.java
IngridPuppet: No need to keep legacy code for now. If this can be removed, please do it.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/OID4VPProfileConfigTest.java
```diff
@@ -49,6 +49,33 @@ void shouldParseConfiguredProfiles() {
...
+        config.setConfig(Map.of(PROFILES_CONFIG, """
+                [
+                  {
+                    "id": "legacy",
+                    "credentials": [
+                      { "id": "primary", "role": "primary", "vct": ["legacy-vct"], "claims": ["sub", "username"] }
+                    ]
+                  }
+                ]
+                """));
```
IngridPuppet: This is a new feature. No need to keep dropped ideas as legacy.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointTest.java
```diff
@@ -54,6 +54,7 @@ public class OID4VPUserAuthEndpointTest extends OID4VPBaseUserAuthEndpointTest {
 
     public static final String VCT_CONFIG_ALT = "https://example.com/vct-alt";
     public static final String DUAL_PROFILE_ID = "dual";
+    // Imported realm fixture ID for the SD-JWT authenticator config used by these integration tests.
     private static final String TEST_REALM_SD_JWT_AUTH_CONFIG_ID = "81d62be9-cf06-4718-8837-fdfb4727b20a";
```
IngridPuppet: In the test vector, change the UUID to "sdjwt-auth-config-id".

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthenticator.java
```diff
@@ -61,6 +65,11 @@ public class SdJwtAuthenticator implements Authenticator {
      */
     public static final String SDJWT_TOKEN_KEY = "sdjwt_token";
 
+    /**
+     * Optional serialized map of DCQL credential IDs to presented SD-JWT VP tokens.
+     */
+    public static final String SDJWT_TOKENS_KEY = "sdjwt_tokens";
+
```
IngridPuppet, line 72: Any reason why these are sent separately?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/OID4VPProfileConfig.java
```diff
@@ -0,0 +1,137 @@
...
+                ? List.of(defaultProfile(context, authConfig))
+                : parseProfiles(configuredProfiles);
+        validateProfiles(this.profiles);
+    }
+
+    public List<AuthenticationProfile> getProfiles() {
+        return profiles;
+    }
+
+    public AuthenticationProfile getProfile(String profileId) {
```
Awambeng, line 39: This method falls back to a default profile lookup when a profile is not found, but this default profile is constructed from the legacy  `VCT_CONFIG`  settings. If a user configures explicit profiles but makes a typo in the  profileId, they'll silently get the default (legacy) behavior instead of an error. This could lead to confusing failures where the wrong credential type is requested.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthenticator.java
```diff
@@ -135,8 +164,37 @@ public void action(AuthenticationFlowContext context) {
...
+        try {
+            AuthorizationContext authContext =
+                    new AuthenticationSessionStore(context.getAuthenticationSession()).getAuthorizationContext();
+            return new OID4VPProfileConfig(context.getSession().getContext(), context.getAuthenticatorConfig())
+                    .getProfile(authContext.getProfileId());
+        } catch (IllegalArgumentException e) {
+            return new OID4VPProfileConfig(context.getSession().getContext(), context.getAuthenticatorConfig())
+                    .getProfile(AuthenticationProfile.DEFAULT_PROFILE_ID);
+        }
+    }
```
Awambeng, line 186: Catching all `IllegalArgumentException's` here may be too broad, as it includes legitimate failure cases such as missing authorization context or invalid profile IDs. This could result in silent fallback to default credentials. It would be safer to handle only expected cases explicitly.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthRequirements.java
```diff
@@ -79,19 +82,23 @@ public SdJwtAuthRequirements(KeycloakContext context, AuthenticatorConfigModel a
...
+        this.expectedVctsPattern = buildExpectedVctsPattern(expectedVcts);
+    }
+
+    public SdJwtAuthRequirements(
+            KeycloakContext context, AuthenticatorConfigModel authConfig, CredentialRequirement credentialRequirement) {
+        this(context, authConfig);
+        this.expectedVcts = credentialRequirement.getCredentialTypes();
+        this.requiredClaims = credentialRequirement.getClaims();
+        this.expectedVctsPattern = buildExpectedVctsPattern(expectedVcts);
     }
```
Awambeng, line 94: This pattern mutates fields after construction, which is technically valid but fragile since it relies on later initialization before `getPresentationRequirements()` is used. This can lead to inconsistent object state. A static factory method would be a safer alternative.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtSupportingCredentialVerifier.java
```diff
@@ -0,0 +1,133 @@
...
+            String sdJwtVpToken,
+            AuthenticatorConfigModel authConfig,
+            String nonce,
+            String aud,
+            boolean requireCryptographicHolderBinding)
+            throws VerificationException {
+        if (!credential.isSelfTrusted()) {
+            throw new VerificationException(
+                    "Credential '%s' uses an unsupported trust policy".formatted(credential.getId()));
+        }
```
Awambeng, line 72: Since `x5c` and anchor-based trust policies aren’t implemented yet, consider adding TODOs to track future support. WDYT?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -152,59 +201,34 @@ private String extractSdJwtVpToken(
...
-            String detailed = "Presented vp_token map does not match DCQL credential query";
-            throw failWithHttpException(
-                    ProcessingError.INVALID_VP_TOKEN,
-                    "Invalid vp_token",
-                    detailed,
-                    Response.Status.BAD_REQUEST,
-                    authContext,
-                    store);
-        }
+    private DcqlQuery singleCredentialQuery(DcqlQuery dcqlQuery, String credentialId) {
```
Awambeng, line 207: `AuthorizationResponseService.singleCredentialQuery()` duplicates `DcqlQueryBuilder.singleCredentialQuery()`. We now have two methods performing essentially the same logic: one in the dcql package (static utility) and one in `AuthorizationResponseService` that extracts a single credential from a query. Consider reusing the builder utility to avoid duplication and keep the logic centralized.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthenticator.java
Ogenbertrand: Hello @IngridPuppet,
They are separate because `SDJWT_TOKEN_KEY` preserves the existing single-primary-credential authenticator path, while `SDJWT_TOKENS_KEY` carries the full DCQL credential-id map only for multi-credential supporting verification.

Awambeng (changes requested):

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/OID4VPProfileConfig.java
Ogenbertrand: `getProfile` only maps blank profile IDs to default, unknown non-blank profile IDs already throw.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthenticator.java
Ogenbertrand: fixed.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthRequirements.java
Ogenbertrand: The constructor is already safe because the object isn’t used before it’s fully built. Changing it would only make the code look different, not safer, and would create unnecessary review work. Please double check again

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
Ogenbertrand: fixed

Awambeng (approved):

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthenticator.java
IngridPuppet: This will be confusing for new contributors. Please could you open a ticket to look deeper into this point? It should be possible to preserve the single-credential flow while accepting a map of tokens with `SDJWT_TOKENS_KEY`. I bet there is a reliable way to identify the primary credential in the map. In general, it's problematic to communicate the same information twice. For now, I'll approve the PR to expedite it. LGTM.

IngridPuppet (approved):

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthenticator.java
Ogenbertrand: I'll open an issue an handle it.

Ogenbertrand:
Thanks @IngridPuppet , @Awambeng for the review.

PR #111, opened 2026-06-17
refactor(oid4vp): Use sdjwt_tokens as sole SD-JWT auth note

### LINKED ISSUE #108: Refactor SD-JWT token handling to avoid duplicating primary credential across `SDJWT_TOKEN_KEY` and `SDJWT_TOKENS_KEY`
In the `SdjwtAuthenticator` class currently, the authentication flow stores the primary SD-JWT credential separately under:
```java
SDJWT_TOKEN_KEY = "sdjwt_token"
```
while supporting or multi-credential presentations are stored as a serialized DCQL credential-id map under:
```java
SDJWT_TOKENS_KEY = "sdjwt_tokens"
```
This design preserves backward compatibility with the existing single-primary-credential authentication flow, but it introduces duplication of credential state and may create confusion for future contributors.

**Problem:**

The same credential information may effectively be communicated twice:

- once as the standalone primary token (SDJWT_TOKEN_KEY)
- once inside the DCQL token map (SDJWT_TOKENS_KEY)

**Goal**

Investigate whether the authentication flow can be simplified by relying solely on:
```java
SDJWT_TOKENS_KEY
```
for both single-credential and multi-credential presentations.

mbunwe-victor:
## Summary

Closes #108 by removing the redundant `sdjwt_token` auth note and using `sdjwt_tokens` as the single store for presented SD-JWT VPs (single- and multi-credential flows).

The authenticator resolves the primary credential from the map by profile credential id. Supporting credentials are passed to the verifier as a filtered subset (excluding the primary). The authorization context is stored on the ephemeral processor session via `AuthenticationSessionStore.storeAuthorizationContext()` so profile resolution works correctly.

## Test plan

- [x] `mvn test` (161 tests)

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthenticator.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthenticator.java
@@ -61,14 +61,11 @@ public class SdJwtAuthenticator implements Authenticator {
     public static final String CHALLENGE_AUD_KEY = "aud";
 
     /**
-     * The authenticating party presents a non-replayable SD-JWT token for authentication.
-     */
-    public static final String SDJWT_TOKEN_KEY = "sdjwt_token";
-
-    /**
-     * Optional serialized map of DCQL credential IDs to presented SD-JWT VP tokens.
+     * Serialized map of DCQL credential IDs to presented SD-JWT VP tokens.
      */
     public static final String SDJWT_TOKENS_KEY = "sdjwt_tokens";
+
+    public static final String PROFILE_ID_KEY = "oid4vp_profile_id";
 
     public static final String REQUIRE_CRYPTOGRAPHIC_HOLDER_BINDING_KEY = "require_cryptographic_holder_binding";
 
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthenticator.java
```diff
@@ -174,9 +178,16 @@ private SdJwtAuthRequirements getAuthenticationRequirements(
...
     private AuthenticationProfile getAuthenticationProfile(AuthenticationFlowContext context) {
+        AuthenticationSessionModel authSession = context.getAuthenticationSession();
         OID4VPProfileConfig profileConfig =
                 new OID4VPProfileConfig(context.getSession().getContext(), context.getAuthenticatorConfig());
-        AuthenticationSessionStore store = new AuthenticationSessionStore(context.getAuthenticationSession());
+
+        String profileId = authSession.getAuthNote(PROFILE_ID_KEY);
+        if (!StringUtil.isBlank(profileId)) {
+            return profileConfig.getProfile(profileId);
+        }
```
IngridPuppet: Look a bit further. Is this new `PROFILE_ID_KEY` value any different from `authContext.getProfileId()`?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/SdJwtAuthenticator.java
IngridPuppet, line 148: At this point, does `sdJwtVpTokens` only consist of "supporting" credentials? That's the intent, right?

IngridPuppet (approved):
LGTM

PR #113, opened 2026-06-29
Implement core logic for verifying mDoc device respones

### LINKED ISSUE #112: Plugin: Implement core logic for verifying mDoc presentations
Not OpenID4VP, just core logic for verifying mDoc presentations. 

Following the model of https://github.com/keycloak/keycloak/tree/main/core/src/main/java/org/keycloak/sdjwt/consumer.

ETA: 3 days

IngridPuppet:
Closes #112


pom.xml
```diff
@@ -128,6 +128,16 @@
...
+        <dependency>
+            <groupId>com.authlete</groupId>
+            <artifactId>cbor</artifactId>
+            <version>1.21</version>
+        </dependency>
+        <dependency>
+            <groupId>com.networknt</groupId>
+            <artifactId>json-schema-validator</artifactId>
+            <version>2.0.2</version>
+        </dependency>
```
Awambeng, line 140: Could you please scope this dependency?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtil.java
```diff
@@ -21,26 +22,41 @@
...
+        try {
             CertificateFactory cf = CertificateFactory.getInstance("X.509");
-            List<X509Certificate> certs = new ArrayList<>();
-            for (String base64Cert : x5c) {
+            for (String base64Cert : base64Certs) {
                 byte[] bytes = Base64.getDecoder().decode(base64Cert);
                 certs.add((X509Certificate) cf.generateCertificate(new ByteArrayInputStream(bytes)));
             }
+        } catch (CertificateException e) {
+            throw new VerificationException("Failed to parse X.5C certificate", e);
```
Awambeng:
```suggestion
            throw new VerificationException("Failed to parse X.509 certificate", e);
```

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/CborUtil.java
```diff
@@ -0,0 +1,148 @@
...
+
+            // Return primitives (Strings, Integers, Booleans, normal ByteArrays) as-is
+            default -> item;
+        };
+    }
+
+    /**
+     * Extracts X5C chain from COSESign1.
+     */
+    public static List<X509Certificate> extractX5Chain(COSESign1 sign1) {
```
Awambeng: Since this method returns `X509Certificate` objects rather than raw `x5c` values, should we rename it to `extractX509Chain()` for clarity?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocParser.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private MdocParser() {}
+
+    public static CBORPairList parseBase64Url(String base64Url) throws MdocEncodingException {
+        if (StringUtil.isBlank(base64Url)) {
+            throw new MdocEncodingException("Input string is null or blank");
+        }
+
+        try {
+            byte[] decoded = Base64.getUrlDecoder().decode(base64Url.trim());
```
Awambeng, line 40: Isn’t this `trim()` redundant, since we already handle whitespace with `StringUtil.isBlank(base64Url)`?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
```diff
@@ -0,0 +1,351 @@
...
+        // Verify response status BAE
+        verifyResponseStatus(mdoc);
+
+        // Verify issuer signature over Mobile Security Objects (MSO)
+        CBORPairList document = extractDocument(mdoc);
+        COSESign1 issuerAuth = extractIssuerAuth(document);
+        verifyIssuerSignature(issuerAuth, truststoreProvider);
+
+        // Verify device key binding
+        CBORPairList mso = (CBORPairList) CborUtil.unwrap(issuerAuth.getPayload());
```
Awambeng, line 83: If the payload doesn’t unwrap to a `CBORPairList`, this will throw a `ClassCastException` instead of a `VerificationException`. Might be worth using `instanceof` or a try-catch to handle it more safely.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
```diff
@@ -0,0 +1,351 @@
...
+        if (presentationRequirements != null) {
+            presentationRequirements.checkIfSatisfiedBy(claims.toJson());
+        }
+    }
+
+    /**
+     * Verify device response status.
+     */
+    private static void verifyResponseStatus(CBORPairList mdoc) throws VerificationException {
+        var status = (CBORInteger) mdoc.findByKey(MdocConstants.L_STATUS).getValue();
```
Awambeng, line 103: If the "status" is missing in the CBOR map, `findByKey()` can return `null`, causing a potential NPE. Consider adding a null check for safety.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationTest.java
```diff
@@ -0,0 +1,247 @@
...
+                new StaticTruststoreProvider(getIssuerCertRef1()),
+                "Device key binding verification failed: missing device signature");
+    }
+
+    /**
+     * Asserts that verifying {@code mdoc} raises a {@link VerificationException} whose message
+     * contains {@code expectedMessageFragment}. When {@code expectedCauseMessageFragment} is
+     * non-null, also asserts that the cause's message contains it.
+     */
+    private static void verifyFails(
```
Awambeng, line 250: `verifyFails` may cause an NPE if `exception.getCause()` is null. Consider making this check more defensive.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationOpts.java
```diff
@@ -0,0 +1,166 @@
...
+    private final String responseUri;
+
+    /**
+     * For OpenID4VP handover - The SHA-256 Thumbprint (as defined in [RFC7638]) of the JWK
+     * advertised in the OpenID4VP request object for response encryption. If none, null.
+     */
+    private final byte[] jwkThumbprint;
+
+    private MdocVerificationOpts(
+            List<Predicate<ObjectNode>> headerVerifiers,
```
Awambeng: This parameter is not used (builder always passes List.of()). Consider removing it or wiring it through the builder.

Awambeng (changes requested):

pom.xml
IngridPuppet: This is intended to be added as a standard dependency.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtil.java
IngridPuppet: Done

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/CborUtil.java
IngridPuppet: Done

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocParser.java
IngridPuppet: The blank check would catch `"  "` but still let `" abc "` through.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
IngridPuppet: The entire CBOR tree was already validated with `MdocParser.parseBase64Url(mdoc)` in the constructor, so it is a certitude that will unwrap to a `CBORPairList`.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
IngridPuppet: The entire CBOR tree was already validated with `MdocParser.parseBase64Url(mdoc)` in the constructor, so it is a certitude the status field is present.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationTest.java
IngridPuppet: Added an a non-null assertion.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtil.java
Ogenbertrand, line 1: I am reviewing this PR against the final version of the OID4VP plugin.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationOpts.java
IngridPuppet: I agree. Removed. The constructor now directly passes the empty list to the parent constructor. I don't think we'll ever need `headerVerifiers` for mDoc verification.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
```diff
@@ -0,0 +1,351 @@
...
+     * Verify device key binding
+     */
+    private void verifyDeviceKeyBinding(CBORPairList document, CBORPairList mso, MdocVerificationOpts opts)
+            throws VerificationException {
+        COSESign1 deviceSignature = extractDeviceSignature(document);
+        COSEKey deviceKey = extractDeviceKey(mso);
+
+        try {
+            // First attempt verification with session transcript computed as per OpenID4VP spec
+            CBORItemList sessionTranscript = OID4VPSessionTranscript.computeSessionTranscript_OID4VPSpec(opts);
```
Ogenbertrand: This firstly verifies with the OpenID4VP transcript, then silently retries with an ISO transcript at line 147. For OpenID4VP redirect invocation, final 1.0 defines the OpenID4VPHandover transcript and says it MUST be used. Accepting a second transcript shape by fallback makes the accepted security context ambiguous and can hide verifier/wallet bugs. This should be selected explicitly by protocol/profile, not “try both”.

https://openid.net/specs/openid-4-verifiable-presentations-1_0-final.html#appendix-B.2.6.1

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/OID4VPSessionTranscript.java
```diff
@@ -0,0 +1,77 @@
...
+    public static CBORItemList computeSessionTranscript_OID4VPSpec(MdocVerificationOpts opts) {
+        byte[] thumbprint = opts.getJwkThumbprint();
+        CBORItem jwkThumbprintCbor =
+                thumbprint != null && thumbprint.length > 0 ? new CBORByteArray(thumbprint) : CBORNull.INSTANCE;
+
+        CBORItemList handoverInfo = new CBORItemList(
+                toCBORString(opts.getClientId()),
+                toCBORString(opts.getOid4vpNonce()),
+                jwkThumbprintCbor,
+                toCBORString(opts.getResponseUri()));
```
Ogenbertrand: This builds OpenID4VPHandoverInfo from clientId, nonce, jwkThumbprint, and responseUri, but `toCBORString` method converts missing values to CBOR null. Th spec requires client_id, nonce, and either redirect_uri or response_uri depending on response mode. Missing required values should fail before computing the transcript: https://openid.net/specs/openid-4-verifiable-presentations-1_0-final.html#appendix-B.2.6.1

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
```diff
@@ -0,0 +1,351 @@
...
+                } catch (IOException e) {
+                    throw new VerificationException(
+                            String.format("Failed to parse value of element with digestId=%s", digestId), e);
+                }
+            }
+            nsClaims.put(namespaceKey, claims);
+        }
+
+        // Bubble up collected claims
+        String docType = extractDocType(mso).getValue();
```
Ogenbertrand, line 265: docType here is not enforced as a first-class mdoc requirement, for `mso_mdoc`, `doctype_value` is required by the OID4VP final DCQL format-specific rules. This should not depend on generic caller-provided PresentationRequirements.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
```diff
@@ -0,0 +1,351 @@
...
+    private static CBORPairList extractDocument(CBORPairList root) throws VerificationException {
+        var documents = (CBORItemList) root.findByKey(MdocConstants.L_DOCUMENTS).getValue();
+        if (documents.getItems().size() != 1) {
+            // A single document is expected.
+            // Request for multiple documents are satisfiable by the OpenID4VP response map.
+            throw new VerificationException(String.format(
+                    "Expected 1 document but received %d", documents.getItems().size()));
+        }
+
+        return (CBORPairList) documents.getItems().getFirst();
```
Ogenbertrand, line 320: The parser scheme only requires `version` and `status` at the root mdoc-device-response-schema.json (line 6) but then this verification later assumes `document` exists.

A status-OK response without documents can likely throw NullPointerException / ClassCastException instead of VerificationException. Since this parses wallet-controlled input, shape failures should be normalized into verification failures, not server errors.

Ogenbertrand (changes requested):
Thanks for the PR @IngridPuppet.
I left some few comments, could you check them ?.
Overall, the crypto core is a solid start.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
IngridPuppet: Fallback must now be explicitly opted in. Thanks.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/OID4VPSessionTranscript.java
IngridPuppet: The code was updated to throw exceptions on required but null data.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
IngridPuppet: The PresentationRequirements interface offers a much more flexible means for the caller to enforce this requirement. Also, it is worth noting the consistency with SD-JWT verification.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
IngridPuppet: Good catch, I agree. A missing "documents" field is now caught properly.

IngridPuppet:
The bot didn't hesitate to cheat from your comments. Very stupid 😅

Awambeng (approved):
LGTM!

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
```diff
@@ -277,7 +304,12 @@ private void verifyValidityInfo(CBORPairList mso, MdocVerificationOpts opts) thr
     }
 
     private static CBORPairList extractDocument(CBORPairList root) throws VerificationException {
-        var documents = (CBORItemList) root.findByKey(MdocConstants.L_DOCUMENTS).getValue();
+        var documentsEntry = root.findByKey(MdocConstants.L_DOCUMENTS);
+        if (documentsEntry == null) {
+            throw new VerificationException("mDoc response is missing the 'documents' field");
+        }
+
+        var documents = (CBORItemList) documentsEntry.getValue();
```
Ogenbertrand, line 312: I’d either make the schema stricter for `status 0` responses or avoid presenting it as full DeviceResponse schema validation.

IngridPuppet: Device responses with a status other than `0` remain structurally valid. The field can take different values, that's why it's there. Consequently, it is okay to refer to the validation as full DeviceResponse schema validation.

Ogenbertrand (approved):
Thanks for considering my comments.
PR LGTM now.

PR #116, opened 2026-07-13
Add support for requesting and accepting presentations of mDoc credentials

### LINKED ISSUE #88: Plugin: Add support for requesting and accepting presentations of mDoc credentials
The plugin currently only supports requesting and verifying SD-JWT presentations. Support for mDoc presentation and verification should be added.

In scope is only what is already possible with SD-JWT, nothing more.

There are plans to push the core code for mDoc verification to Keycloak. Find that code [here](https://github.com/ba-itsys/keycloak-extension-oid4vp). But that's not a priority for the code owner atm. For the time being, consider adding a dedicated package to the plugin for the same logic.

It's critical that key binding and non-replay are enforced.

ETA: 3 days

IngridPuppet:
Closes #88 

Main updates in the PR:
- Introduce an OpenID4VP authenticator (to supersede SD-JWT authenticator)
- Reorganize configuration across VerifierConfig and SdJwt/Mdoc AuthRequirements classes
- Refactor SdJwt verification into SdJwtCredentialVerifier then add MdocCredentialVerifier for the verification of mDoc presented tokens
- Fix DCQL query construction and authorization response flow to not only target SD-JWT

Planned follow-up issues:
- https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/117
- https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/118
- https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/119
- https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/120
- https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/121
- https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/122
- https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/123
- https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/124
- https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/125


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocAuthRequirements.java
```diff
@@ -0,0 +1,57 @@
...
+        return MdocVerificationOpts.builder()
+                .withClientId(clientId)
+                .withOid4vpNonce(oid4vpNonce)
+                .withResponseUri(responseUri)
+                .withJwkThumbprint(jwkThumbprint)
+                .withMdocGeneratedNonce(mdocGeneratedNonce)
+                .withFallbackToIsoSpecSessionTranscript(authRequirements.shouldFallbackToIsoSpecSessionTranscript())
+                .withAllowedMaxAge(authRequirements.getHolderBindingProofMaxAge())
+                .build();
+    }
```
Ogenbertrand: I don’t think we should apply holderBindingProofMaxAge to mDoc MSO validityInfo.signed. That timestamp is the issuer/MSO signing time, not the wallet presentation time. For mDoc, replay protection should come from the device signature over the OpenID4VP SessionTranscript, which already includes the verifier nonce/handover values. This could reject valid long-lived mDocs after the default 60s.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/sdjwt/SdJwtCredentialVerifier.java
```diff
@@ -0,0 +1,141 @@
...
+                        String.format(
+                                "Token status verification failed for credential to requirement '%s'",
+                                credential.getId()),
+                        e);
+            }
+        }
+
+        if (credential.isPrimary()) {
+            validateTransactionData(context.getAuthenticationSession(), token);
+        }
```
Ogenbertrand, line 97: transaction_data validation currently only exists in the SD-JWT verifier. If the primary credential is mDoc and transaction_data was requested, the flow accepts the presentation without enforcing that binding. Can we either implement the mDoc equivalent or reject transaction_data when the primary credential format is mso_mdoc?

Ogenbertrand (changes requested):
Thanks for this massive PR @IngridPuppet.
I just have some minor comments for you.

In the DcqlQueryValidator class at line 44, DCQL validation still treats every non-SD-JWT format as unsupported, including mso_mdoc. I guess you didn't notice while implementing.

I am facing some failing tests is it only me:

Image

IngridPuppet:
> In the DcqlQueryValidator class at line 44, DCQL validation still treats every non-SD-JWT format as unsupported, including mso_mdoc. I guess you didn't notice while implementing.

Yes, I decided to handle that separately in #117. I believe DCQL validation for mDoc would entail some specific code. This PR is already tooo big.

> I am facing some failing tests is it only me:

It's now required to compile the library before running tests because it has some dependencies not available in Keycloak now. Testcontainers need a "shaded" jar. I assume you ran `./mvnw package`. Please try the following command instead:

```bash
./mvnw package -DskipTests && ./mvnw test
```

It'll be good to find a better solution later.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocAuthRequirements.java
IngridPuppet: This is verrry critical. You're right I've assumed that that was presentation time. I'll check a bit then take it off.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/sdjwt/SdJwtCredentialVerifier.java
IngridPuppet: Thanks. I've intentionally not cared about that for mso_mdoc for now due to how big the PR already is. I'll explicitly reject it, and open a follow-up ticket for that.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * {@link CredentialVerifier} implementation backing verification of {@code mso_mdoc}
+ * (ISO/IEC 18013-5) credentials presented via OpenID4VP.
+ *
+ * <p>Owns the mDoc-specific verification entry points (issuer signature, device key
+ * binding, MSO digest integrity, validity info, claim requirements, trust policy).
+ */
+public class MdocCredentialVerifier implements CredentialVerifier {
+
+    private static final Logger logger = Logger.getLogger(MdocCredentialVerifier.class);
```
Ogenbertrand, line 35: logger is unused in this class. Can we remove it to keep the verifier clean?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
```diff
@@ -0,0 +1,104 @@
...
+        String mdocGeneratedNonce = authorizationContext.getResponseApuNonce();
+
+        MdocVerificationOpts opts = authReqs.getMdocVerificationOpts(
+                requestObject.getClientId(),
+                requestObject.getNonce(),
+                requestObject.getResponseUri(),
+                jwkThumbprint,
+                mdocGeneratedNonce);
+
+        AtomicReference<JsonNode> payloadRef = new AtomicReference<>();
```
Ogenbertrand, line 67: This AtomicReference makes the flow harder to follow and assumes `verifyPresentation()` always invokes the callback exactly once. Could we make `MdocVerificationContext.verifyPresentation()` return the verified payload/claims, then run the presentation requirements explicitly? At minimum, please guard `payloadRef.get()` before dereferencing.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
```diff
@@ -0,0 +1,104 @@
...
+        PresentationRequirements requirements = payload -> {
+            payloadRef.set(payload);
+            authReqs.getPresentationRequirements().checkIfSatisfiedBy(payload);
+        };
+
+        new MdocVerificationContext(token).verifyPresentation(opts, requirements, truststore);
+
+        // TODO: Implement revocation status check for mDoc credentials
+        // if (authReqs.shouldEnforceRevocationStatus()) {}
```
Ogenbertrand: If revocation enforcement is enabled, mDoc verification currently still succeeds silently. Since mDoc revocation is not implemented yet, can we fail fast when `authReqs.shouldEnforceRevocationStatus()` is true instead of leaving this as a TODO after successful verification?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
```diff
@@ -0,0 +1,104 @@
...
+        return Optional.ofNullable(requestObject.getClientMetadata())
+                .map(ClientMetadata::getJwks)
+                .map(JSONWebKeySet::getKeys)
+                .filter(keys -> keys.length > 0)
+                .map(keys -> keys[0])
+                .map(JWKSUtils::computeThumbprint)
+                .map(Base64Url::decode)
+                .orElse(null);
+    }
+
```
Ogenbertrand, line 99: `computeJwkThumbprint()` is handover/session-transcript support, not really credential verification. Could we move it closer to OID4VPSessionTranscript or a small mDoc handover utility so MdocCredentialVerifier stays focused on orchestration?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocAuthRequirements.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Configured presentation requirements on mDoc device responses for authentication.
+ */
+public class MdocAuthRequirements {
+
+    private static final Logger logger = Logger.getLogger(MdocAuthRequirements.class);
+
+    private final AuthRequirements authRequirements;
+
+    private List<String> expectedDocTypes;
```
Ogenbertrand, line 21: Small cleanup: `expectedDocTypes` and `requiredClaims` are mutated after constructor delegation. will it be safe to make it immutable by assigning these fields once through constructors?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/VerifierConfig.java
```diff
@@ -152,17 +142,35 @@ private static X509Certificate validateX5CCertificate(String certificate) {
...
+                    "transactionData cannot be used when requireCryptographicHolderBinding is false (OpenID4VP B.3.3)");
+        }
+
+        // TODO: Implement transaction_data binding verification for mso_mdoc credentials
+        // (follow-up ticket). Until then, transaction_data must not be requested when the
+        // primary credential of any profile is mso_mdoc — the mDoc equivalent of KB-JWT
+        // transaction_data_hashes binding is not yet enforced.
+        if (anyProfileHasMdocPrimary(profileConfig)) {
+            throw new IllegalStateException("transactionData is not yet supported for mso_mdoc primary credentials");
+        }
```
Ogenbertrand, line 164: This rejects transactionData when any configured profile has an mDoc primary, even if the selected profile is SD-JWT. Since transaction data is applied to the selected primary credential, can this validation move to where the selected `AuthenticationProfile` is known?

Ogenbertrand (commented):
I have a few more comments @IngridPuppet, Could you check them.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
IngridPuppet: It's generally fine for loggers. It'll soon be used.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
IngridPuppet: > Could we make MdocVerificationContext.verifyPresentation() return the verified payload/claims, then run the presentation requirements explicitly?

This is generally not a good pattern. There are multiple things you would want to return from that method for convenience. Notice that the signature aligns with the SdJwtVerificationContext class in Keycloak.

> At minimum, please guard payloadRef.get() before dereferencing.

Verification passed, so it's there. That's the guard.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointFactory.java
```diff
@@ -67,11 +68,25 @@ public void postInit(KeycloakSessionFactory factory) {
...
+                .map(AuthenticationExecutionModel::getAuthenticator)
+                .anyMatch(OID4VPAuthenticatorFactory.PROVIDER_ID::equals);
+
+        if (!hasExpectedAuthenticator) {
+            throw new IllegalStateException(String.format(
+                    "Authentication flow '%s' already exists in realm '%s' but does not contain the expected "
+                            + "OpenID4VP authenticator '%s'. Refusing to start. Consider backing up the current "
+                            + "flow configuration and removing it so Keycloak can create a new one on next restart.",
+                    OID4VP_AUTH_FLOW, realm.getName(), OID4VPAuthenticatorFactory.PROVIDER_ID));
         }
```
forkimenjeckayang, line 90: This rejects the normal upgrade state created by the current release. `main` creates the same `oid4vp auth` flow with provider ID `sd-jwt-authenticator`; this PR removes that provider and registers `oid4vp-authenticator`, so every existing plugin-managed realm reaches this branch during `PostMigrationEvent` and can prevent Keycloak startup. Deleting the flow as suggested also discards its authenticator configuration, while this PR renames persisted keys such as `vct` to `credentialTypes` and `kbJwtMaxAge` to `holderBindingProofMaxAge`. 

Could we  migrate the existing execution and its config in place, or retain a compatibility provider that performs the migration?

That being the case we can add an upgrade test that starts with a realm export from `main`, verifies startup succeeds, and verifies the previous configuration values are preserved under the new authenticator.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
IngridPuppet: Notice that it's already a delegate. The actual logic already lives in `JWKSUtils::computeThumbprint`. Most lines just parse the request object, which is neither a concern of `OID4VPSessionTranscript`.

pom.xml
```diff
@@ -212,6 +215,34 @@
...
+                            <filters>
+                                <filter>
+                                    <artifact>*:*</artifact>
+                                    <excludes>
+                                        <exclude>META-INF/*.SF</exclude>
+                                        <exclude>META-INF/*.DSA</exclude>
+                                        <exclude>META-INF/*.RSA</exclude>
+                                    </excludes>
+                                </filter>
+                            </filters>
```
forkimenjeckayang, line 241: Attaching the shaded artifact leaves the normal `keycloak-oid4vp-plugin-{version}.jar` as a thin JAR. That is still the artifact named by the README and root `docker-compose.yml`, but it contains neither the new Authlete CBOR classes nor NetworkNT schema classes, so the documented deployment will fail when those paths load. The test harness avoids this by mounting `-all.jar`. The `-all.jar` is not a clean alternative yet either: the current unbounded shade includes Keycloak-provided framework packages, including 1,169 Jackson and 59 SLF4J class files, which risks split packages and class-version conflicts in Keycloak. 

Could we produce one well-defined deployable provider artifact: include only dependencies Keycloak does not provide, and exclude or relocate framework dependencies supplied by Keycloak. Then update README, compose/demo selection, release publishing, and Testcontainers to exercise that same artifact. A clean `./mvnw test` should also build or resolve its test provider automatically rather than depending on a prior package invocation ?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
```diff
@@ -0,0 +1,384 @@
...
+                        case BindingRule.CLAIM_EQUALS_USER_ATTRIBUTE ->
+                            readUserAttribute(user, rule.getUserAttribute());
+                        default ->
+                            throw new IllegalStateException(
+                                    String.format("Unsupported binding rule type: %s", rule.getType()));
+                    };
+
+            if (!Objects.equals(expectedValue, supportingValue)) {
+                throw new VerificationException(String.format(
+                        "Supporting credential '%s' failed binding rule '%s'", credential.getId(), rule.getType()));
```
forkimenjeckayang, line 220: This regresses the previous supporting-credential binding check. The old `SdJwtSupportingCredentialVerifier` rejected a blank supporting value before comparing it; `Objects.equals()` now accepts `null == null` and `"" == ""`. Profile validation does not ensure that a binding rule's credential claim is requested/present or that its primary claim/user attribute has a value. For mDoc specifically, `SimpleMdocPresentationDefinition` also treats an explicit CBOR/JSON null node as present, while `MdocCredentialVerifier.readClaim()` converts it to `null`. A supporting credential can therefore pass a configured binding rule without carrying a binding value whenever the expected claim or user attribute is also absent. 

Can we please  require both values to be non-blank before equality, validate binding-rule references when profiles are parsed, and add regression tests for missing, null, and blank values across mDoc and SD-JWT supporting credentials.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocAuthRequirements.java
IngridPuppet: I've had this AI suggestion excessively. It unnecessaritly makes the code more complicated.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
IngridPuppet: Good suggestion. Done.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/VerifierConfig.java
IngridPuppet: All configured profiles are potentially selectable. I'd argue it is wanted to fail fast.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/TrustedProviderResolver.java
```diff
@@ -0,0 +1,105 @@
...
+
+        List<X509Certificate> certs = new ArrayList<>();
+        List<String> invalid = new ArrayList<>();
+        for (String anchor : trust.getAnchors()) {
+            try {
+                certs.add(CertificateUtil.parseCertificate(anchor));
+            } catch (CertificateException e) {
+                invalid.add(anchor);
+                logger.warnf(e, "Failed to parse x5c trust anchor: (credential '%s') %s", credentialId, anchor);
+            }
```
forkimenjeckayang: `CertificateUtil.parseCertificate()` first calls `Base64.getDecoder().decode()`, which throws `IllegalArgumentException` for malformed Base64 rather than `CertificateException`. This new resolver catches only `CertificateException`, and the primary `OID4VPAuthenticator` path catches only `VerificationException`, so an invalid configured anchor such as `%%%` escapes the normal credential-rejection path and can turn every login using that profile into an internal server error. 

Could we validate trust policy values while parsing the profile and/or catch both decoding and certificate parsing failures here and wrap them in `VerificationException`?

Then add tests for malformed Base64, invalid DER, null/blank anchors, and a controlled authentication failure rather than an uncaught runtime exception?

forkimenjeckayang (commented):
A few comments @IngridPuppet 
WDYT?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointFactory.java
IngridPuppet: That would definitely make sense if the plugin were used in production by thousands of environments, which is not really the case. Maintaining some migration logic would be unnecessary hassle at this stage I would say. Lots of things are still changing.

pom.xml
IngridPuppet: Thanks. To ease reviewing this point, I've opened a follow up ticket (with this description):
- https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/125

It's definitely unwanted that provided packages are also included in the jar.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
IngridPuppet: This revealed a much more dangerous problem. Required claims with blank values were not caught properly. This is now fixed.

> Can we please require both values to be non-blank before equality

Added, though logically just defensive check.

> validate binding-rule references when profiles are parsed,

Done.

> add regression tests for missing, null, and blank values across mDoc and SD-JWT supporting credentials.

Added for mDoc. DCQL validation, enabled for SD-JWT, already catches these scenarios.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/TrustedProviderResolver.java
IngridPuppet: > Could we validate trust policy values while parsing the profile and/or catch both decoding and certificate parsing failures here and wrap them in `VerificationException`?

Done.

> Then add tests for malformed Base64, invalid DER, null/blank anchors, and a controlled authentication failure rather than an uncaught runtime exception?

These are cases of IllegalStateException's, for which the user shouldn't be blamed. So 500 is fine.

Ogenbertrand (approved):
Thanks for the massive contribution @IngridPuppet.
I have no further comments for this PR.

forkimenjeckayang (approved):
LGTM

Issue #125, opened 2026-07-15
Fix plugin release jar to also maybe include unprovided dependencies


IngridPuppet:
Follow-up to:
- #88 

> Attaching the shaded artifact leaves the normal keycloak-oid4vp-plugin-{version}.jar as a thin JAR. That is still the artifact named by the README and root docker-compose.yml, but it contains neither the new Authlete CBOR classes nor NetworkNT schema classes, so the documented deployment will fail when those paths load. The test harness avoids this by mounting -all.jar. The -all.jar is not a clean alternative yet either: the current unbounded shade includes Keycloak-provided framework packages, including 1,169 Jackson and 59 SLF4J class files, which risks split packages and class-version conflicts in Keycloak.
> 
> Could we produce one well-defined deployable provider artifact: include only dependencies Keycloak does not provide, and exclude or relocate framework dependencies supplied by Keycloak. Then update README, compose/demo selection, release publishing, and Testcontainers to exercise that same artifact. A clean ./mvnw test should also build or resolve its test provider automatically rather than depending on a prior package invocation ?

ETA: 1 day

PR #128, opened 2026-07-17
Docs/national wallet pid verification config

### LINKED ISSUE #129: Test PID verification and document
ETA: 1 day

Ogenbertrand:
- I added a length guard before user-id lookup so externally presented credential subjects cannot trigger expensive or unsafe Keycloak user lookup paths, while still allowing normal username-based recovery.
- Adds production documentation for configuring national wallet PID verification with OpenID4VP, including verifier certificate inputs, trust policy setup, single and multi-credential profile examples, and docs navigation/README links. It also keeps local test material out of git and includes a small hardening change for OID4VP user recovery.

### DIFF
--- a/docs/modules/ROOT/pages/national-wallet-pid-verification-configuration.adoc
+++ b/docs/modules/ROOT/pages/national-wallet-pid-verification-configuration.adoc
@@ -1,0 +1,254 @@
+= National Wallet PID Verification Configuration
+
+This page describes how to configure Keycloak OpenID4VP authentication for presentation flows involving a national wallet PID credential.
+
+The configuration supports both:
+
+* a single-credential login profile;
+* a multi-credential login profile where one credential is used for Keycloak user binding and one or more additional credentials are verified as supporting credentials.
+
+== References
+
+Use the official German EUDI Wallet and EUDI Architecture documentation for ecosystem onboarding, Sandbox access, and certificate lifecycle details:
+
+* https://eudi-wallet.gov.de/en/ecosystem-knowledge-center[German EUDI Wallet Ecosystem Knowledge Center]
+* https://bmi.usercontent.opencode.de/eudi-wallet/developer-guide/sandbox/onboarding/joining/[German Sandbox onboarding guide]
+* https://bmi.usercontent.opencode.de/eudi-wallet/developer-guide/rp/onboarding/registrar_certificate_usage/[Using registrar certificates in presentation requests]
+* https://bmi.usercontent.opencode.de/eudi-wallet/developer-guide/rp/PID_Presentation/[Relying Party PID presentation guide]
+* https://docs.eudi.dev/latest/build/supporting-ecosystem-services/trusted-list-service/[EUDI Trust List Service]
+
+== Overview
+
+For the national wallet PID flow, Keycloak acts as an OpenID4VP verifier.
+
+The recommended authentication model is:
+
+. A primary credential identifies the Keycloak user.
+. A PID credential is requested as a supporting credential.
+. Each credential is verified according to its own format and trust policy.
+. Authentication succeeds only when all required credentials are valid.
+
+The PID can be represented as an mDoc credential with:
+
+[source,text]
+----
+Format: mso_mdoc
+Doctype: eu.europa.ec.eudi.pid.1
+----
+
+== Prerequisites
+
+* Keycloak with this plugin deployed.
+* A public HTTPS Keycloak base URL reachable by the wallet.
+* Access to the German EUDI Ecosystem Sandbox Registrar.
+* Verifier access and registration certificate material from the Sandbox.
+* A wallet containing the credentials required by the selected authentication profile.
+
+== Verifier Certificate Material
+
+The national wallet profile requires verifier certificate material issued or accepted by the German EUDI Sandbox.
+
+[cols="1,2"]
+|===
+|Field |Description
+
+|Access certificate
+|Base64 DER encoded X.509 verifier access certificate.
+
+|Registration certificate
+|Full verifier registration certificate JWT.
+|===
+
+The access certificate must correspond to the Keycloak signing key used for OpenID4VP request objects and must identify the public hostname used by Keycloak.
+
+The registration certificate should describe the verifier purpose and the credential claims requested from the wallet.
+
+For a PID mDoc profile, the registration certificate should cover:
+
+[source,text]
+----
+Format: mso_mdoc
+Doctype: eu.europa.ec.eudi.pid.1
+Claims:
+  eu.europa.ec.eudi.pid.1 / given_name
+  eu.europa.ec.eudi.pid.1 / family_name
+----
+
+== Keycloak Signing Key
+
+Configure an active realm key provider for signing OpenID4VP request objects.
+
+Recommended key provider properties:
+
+[source,text]
+----
+Algorithm: ES256
+Keystore Type: PKCS12
+Key use: sig
+----
+
+The private key in this provider must match the public key represented by the configured access certificate.
+
+== OpenID4VP Authenticator Configuration
+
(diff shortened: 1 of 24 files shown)


pom.xml
```diff
@@ -243,6 +242,27 @@
...
+                            <target>
+                                <copy file="${project.build.directory}/${project.build.finalName}.jar"
+                                      tofile="${project.build.directory}/${project.build.finalName}-all.jar"
+                                      overwrite="true"/>
+                            </target>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
             <plugin>
```
IngridPuppet, line 254: This basically makes the main jar a shaded, bloated jar. Then, it creates a copy of it ending in "-all.jar", so tests continue to pass. This is a very misleading trick that we should not move ahead with. If one flow needs the shaded plugin, let it use the "-all.jar" we previously produced directly. FYI, Ticket #125 is dedicated to this "issue". So feel free to contribute to it with your ideas. This documentation PR should not attempt to change the current release flow.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
```diff
@@ -315,6 +320,10 @@ private UserModel recoverAuthenticatingUser(
         return user;
     }
 
+    private static boolean isSafeUserIdLookupSubject(String subject) {
+        return !StringUtil.isBlank(subject) && subject.length() <= MAX_USER_ID_LOOKUP_SUBJECT_LENGTH;
+    }
+
```
IngridPuppet: This change is paved with good intentions but implemented in an invasive, non-extendable way. 

Could you open a separate issue where the problem is clearly described and motivated? What are chances that this problem propagates from the controller to this authenticator? Is it the only field affected? Is the fix sufficiently tested? 

Right now, I don't think we ran into any pressing issue linkable to this. So the changes should be removed and the PR should only focus on documentation as per its title.

docs/modules/ROOT/pages/national-wallet-pid-verification-configuration.adoc
```diff
@@ -0,0 +1,254 @@
...
+== Overview
+
+For the national wallet PID flow, Keycloak acts as an OpenID4VP verifier.
+
+The recommended authentication model is:
+
+. A primary credential identifies the Keycloak user.
+. A PID credential is requested as a supporting credential.
+. Each credential is verified according to its own format and trust policy.
+. Authentication succeeds only when all required credentials are valid.
```
IngridPuppet:
```suggestion
. Authentication succeeds only when all required credentials are valid and the supporting credential meaningfully bound to the primary credential.
```

docs/modules/ROOT/pages/national-wallet-pid-verification-configuration.adoc
```diff
@@ -0,0 +1,254 @@
...
+* https://bmi.usercontent.opencode.de/eudi-wallet/developer-guide/sandbox/onboarding/joining/[German Sandbox onboarding guide]
+* https://bmi.usercontent.opencode.de/eudi-wallet/developer-guide/rp/onboarding/registrar_certificate_usage/[Using registrar certificates in presentation requests]
+* https://bmi.usercontent.opencode.de/eudi-wallet/developer-guide/rp/PID_Presentation/[Relying Party PID presentation guide]
+* https://docs.eudi.dev/latest/build/supporting-ecosystem-services/trusted-list-service/[EUDI Trust List Service]
+
+== Overview
+
+For the national wallet PID flow, Keycloak acts as an OpenID4VP verifier.
+
+The recommended authentication model is:
```
IngridPuppet: I'm not sure we can answered who issued the recommendation.
```suggestion
The authentication model is:
```

docs/modules/ROOT/pages/national-wallet-pid-verification-configuration.adoc
```diff
@@ -0,0 +1,254 @@
+= National Wallet PID Verification Configuration
+
+This page describes how to configure Keycloak OpenID4VP authentication for presentation flows involving a national wallet PID credential.
```
IngridPuppet: - Revise the title. It's currently misleading. It should be clear that a PID credential can be required in addition to a primary credential. Examples: 
   - Configuring authentication to require the additional presentation of a PID credential
   - Configuring dual-credential authentication for the German wallet
- Prefer "German wallet" to "national wallet". It's not national beyond Germany.

docs/modules/ROOT/pages/national-wallet-pid-verification-configuration.adoc
```diff
@@ -0,0 +1,254 @@
...
+
+[source,json]
+----
+{
+  "type": "eudi_pid_trust_list",
+  "trustListUrl": "https://bmi.usercontent.opencode.de/eudi-wallet/test-trust-lists/pid-provider.jwt",
+  "trustListSigningCertificate": "<base64 DER trust-list signing certificate>",
+  "serviceType": "http://uri.etsi.org/19602/SvcType/PID/Issuance"
+}
+----
```
IngridPuppet: Not many will get where this is supposed to be added if it comes before the profile configuration.

docs/modules/ROOT/pages/national-wallet-pid-verification-configuration.adoc
```diff
@@ -0,0 +1,254 @@
...
+Require Not Before claim (SD-JWT): Off
+Require Expiration claim (SD-JWT): Off
+Verify issuer claim (SD-JWT): On
+Fall back to ISO-spec session transcript (mDoc): Off
+Reject revoked credentials: Configure according to the deployment's status-list policy
+----
+
+Use `direct_post.jwt` when the wallet returns an encrypted authorization response.
+
+The login theme must also support OpenID4VP login buttons. For this plugin, use the `keycloak.v2+oid4vp` login theme.
```
IngridPuppet: Most of this is documented in `oid4vp-user-authentication.adoc`. The ES256 signing key is simpler to add with a usual generated key provider. Please could you only focus on new information to reduce verbosity? It is okay to briefly recall some points though.

docs/modules/ROOT/pages/national-wallet-pid-verification-configuration.adoc
```diff
@@ -0,0 +1,254 @@
...
+          }
+        ]
+      }
+    ]
+  }
+]
+----
+
+A PID-only profile can also be configured if the PID credential contains the claims required by the deployment's user-binding policy.
+
```
IngridPuppet: This is not possible. Only self-issued credentials currently work as primary credentials.

IngridPuppet (changes requested):

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
Ogenbertrand: I opened this ticket: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/133 for it.

docs/modules/ROOT/pages/national-wallet-pid-verification-configuration.adoc
Ogenbertrand: I shortened the document by removing this.
Thanks

docs/modules/ROOT/pages/national-wallet-pid-verification-configuration.adoc
Ogenbertrand: change the wording from “The recommended authentication model is:” to “The authentication model is:” so the doc no longer implies we know who issued the recommendation.

IngridPuppet (approved):
Thank you for the update.

PR #131, opened 2026-07-21
fix(oid4vp): enforce revocation status validation for mDoc presentations

### LINKED ISSUE #118: Check the revocation status of mDoc presentations
Follow up to:
- #88 

As it is done for presented SD-JWTs, enforce that mDoc presentations are not revoked as per Token Status List.

ETA: 1 day

Awambeng:
This PR adds credential status validation for mso_mdoc credentials, bringing mDoc verification in line with the existing SD-JWT VC revocation status validation flow.

When `enforceRevocationStatus` is enabled, the MSO (Mobile Security Object) extracted from the mDoc `issuerAuth` is validated against the configured Token Status List to ensure the presented credential has a valid status.

Key changes:
- Add `StatusListJwtFetcher` and `ReferencedTokenValidator` support to `MdocCredentialVerifier`
- Add `extractMsoPayload()` to extract and convert the issuer-signed MSO payload into JSON format
- Inject `httpFetcher` into `MdocCredentialVerifier` through `OID4VPAuthenticatorFactory`
- Add `MdocRevocationStatusTest` covering missing, valid, and invalid credential status scenarios

Closes #118

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
@@ -13,6 +13,7 @@ import com.authlete.cose.COSEException;
 import com.authlete.cose.COSEKey;
 import com.authlete.cose.COSESign1;
 import com.authlete.cose.COSEVerifier;
+import com.fasterxml.jackson.databind.JsonNode;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.crypto.PKIXVerificationUtil;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.trust.TrustAnchorProvider;
 import java.io.IOException;
@@ -48,6 +49,7 @@ public class MdocVerificationContext {
             List.of(JavaAlgorithm.SHA256, JavaAlgorithm.SHA384, JavaAlgorithm.SHA512);
 
     private final CBORPairList mdoc;
+    private JsonNode verifiedMsoPayload;
 
     public MdocVerificationContext(String mdoc) throws VerificationException {
         try {
@@ -81,6 +83,7 @@ public class MdocVerificationContext {
 
         // Verify device key binding
         CBORPairList mso = (CBORPairList) CborUtil.unwrap(issuerAuth.getPayload());
+        storeVerifiedMsoPayload(mso);
         verifyDeviceKeyBinding(document, mso, opts);
 
         // Verify that presented claims are protected by digests in MSO
@@ -303,7 +306,7 @@ public class MdocVerificationContext {
         }
     }
 
-    public static CBORPairList extractDocument(CBORPairList root) throws VerificationException {
+    static CBORPairList extractDocument(CBORPairList root) throws VerificationException {
         var documentsEntry = root.findByKey(MdocConstants.L_DOCUMENTS);
         if (documentsEntry == null) {
             throw new VerificationException("mDoc response is missing the 'documents' field");
@@ -338,7 +341,7 @@ public class MdocVerificationContext {
                 deviceSigned.findByKey(MdocConstants.L_NAME_SPACES).getValue();
     }
 
-    public static COSESign1 extractIssuerAuth(CBORPairList document) throws VerificationException {
+    static COSESign1 extractIssuerAuth(CBORPairList document) throws VerificationException {
         var issuerSigned =
                 (CBORPairList) document.findByKey(MdocConstants.L_ISSUER_SIGNED).getValue();
         var issuerAuth = (CBORItemList)
@@ -379,5 +382,21 @@ public class MdocVerificationContext {
         } catch (COSEException e) {
             throw new VerificationException("Failure parsing deviceSignature as COSE_Sign1", e);
         }
+    }
+
+    private void storeVerifiedMsoPayload(CBORPairList mso) throws VerificationException {
+        try {
+            this.verifiedMsoPayload = JsonSerialization.mapper.valueToTree(new CBORParser(mso.encode()).next());
+        } catch (IOException e) {
+            throw new VerificationException("Failed to serialize verified MSO", e);
+        }
+    }
+
+    /**
+     * @see <a href="https://learn.mattr.global/docs/holding/credential-claiming-guides/revocation-status-check#how-status-information-is-stored">
+     * MATTR docs — status information in the MSO</a>
+     */
+    public JsonNode getVerifiedMsoPayload() {
+        return verifiedMsoPayload;
     }
 }
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
@@ -83,8 +83,8 @@ public class MdocVerificationContext {
 
         // Verify device key binding
         CBORPairList mso = (CBORPairList) CborUtil.unwrap(issuerAuth.getPayload());
-        storeVerifiedMsoPayload(mso);
         verifyDeviceKeyBinding(document, mso, opts);
+        storeVerifiedMsoPayload(mso);
 
         // Verify that presented claims are protected by digests in MSO
         CBORPairList namespaces = extractNamespaces(document);
@@ -392,10 +392,6 @@ public class MdocVerificationContext {
         }
     }
 
-    /**
-     * @see <a href="https://learn.mattr.global/docs/holding/credential-claiming-guides/revocation-status-check#how-status-information-is-stored">
-     * MATTR docs — status information in the MSO</a>
-     */
     public JsonNode getVerifiedMsoPayload() {
         return verifiedMsoPayload;
     }
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
@@ -303,7 +303,7 @@ public class MdocVerificationContext {
         }
     }
 
-    private static CBORPairList extractDocument(CBORPairList root) throws VerificationException {
+    public static CBORPairList extractDocument(CBORPairList root) throws VerificationException {
         var documentsEntry = root.findByKey(MdocConstants.L_DOCUMENTS);
         if (documentsEntry == null) {
             throw new VerificationException("mDoc response is missing the 'documents' field");
@@ -338,7 +338,7 @@ public class MdocVerificationContext {
                 deviceSigned.findByKey(MdocConstants.L_NAME_SPACES).getValue();
     }
 
-    private static COSESign1 extractIssuerAuth(CBORPairList document) throws VerificationException {
+    public static COSESign1 extractIssuerAuth(CBORPairList document) throws VerificationException {
         var issuerSigned =
                 (CBORPairList) document.findByKey(MdocConstants.L_ISSUER_SIGNED).getValue();
         var issuerAuth = (CBORItemList)
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
@@ -2,11 +2,7 @@ package io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.authenticator.mdoc;
 
 import static io.github.adorsysgis.keycloak.protocol.oid4vc.mdoc.MdocConstants.L_NAME_SPACES;
 
-import com.authlete.cbor.CBORParser;
 import com.fasterxml.jackson.databind.JsonNode;
-import io.github.adorsysgis.keycloak.protocol.oid4vc.mdoc.CborUtil;
-import io.github.adorsysgis.keycloak.protocol.oid4vc.mdoc.MdocEncodingException;
-import io.github.adorsysgis.keycloak.protocol.oid4vc.mdoc.MdocParser;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.mdoc.MdocVerificationContext;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.mdoc.MdocVerificationOpts;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.authenticator.CredentialFormat;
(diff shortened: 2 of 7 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
```diff
@@ -70,17 +88,64 @@ public JsonNode verifyCredential(
...
+        }
+
+        var first = items.getFirst();
+
+        if (!(first instanceof CBORPairList doc)) {
+            throw new VerificationException("mDoc document has invalid format");
+        }
+
+        return doc;
+    }
```
IngridPuppet: This logic exists in `MdocVerificationContext`. Please refactor to reuse code as much as possible.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
```diff
@@ -70,17 +88,64 @@ public JsonNode verifyCredential(
...
 
-        // TODO: Implement revocation status check for mDoc credentials
+        new MdocVerificationContext(token).verifyPresentation(opts, requirements, truststore);
+
         if (authReqs.shouldEnforceRevocationStatus()) {
-            throw new VerificationException(
-                    "Revocation status enforcement is not yet implemented for mso_mdoc credentials");
+            try {
+                JsonNode msoPayload = extractMsoPayload(token);
+                tokenStatusValidator.validate(msoPayload);
```
IngridPuppet: Could you share some reference confirming that the status lives in this payload?

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/stub/CustomOID4VPAuthenticatorFactory.java
```diff
@@ -38,7 +38,7 @@ public Authenticator create(KeycloakSession session) {
         Map<String, CredentialVerifier> handlers = new LinkedHashMap<>();
         StatusListJwtFetcher httpFetcher = new MockTrustedStatusListJwtFetcher(session);
         handlers.put(CredentialFormat.SD_JWT_VC.getValue(), new SdJwtCredentialVerifier(httpFetcher));
-        handlers.put(CredentialFormat.MSO_MDOC.getValue(), new MdocCredentialVerifier());
+        handlers.put(CredentialFormat.MSO_MDOC.getValue(), new MdocCredentialVerifier(httpFetcher));
         return new OID4VPAuthenticator(handlers);
```
IngridPuppet, line 42: Please one integration test if possible.

IngridPuppet (changes requested):

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
```diff
@@ -70,17 +82,42 @@ public JsonNode verifyCredential(
...
```
forkimenjeckayang: > This new mDoc path delegates to `ReferencedTokenValidator`, but that validator accepts the fetched Status List JWT after checking only `typ=statuslist+jwt` and the embedded `status_list`. It does not require `sub` or `iat`, does not verify that `sub` exactly equals the `status_list.uri` from the signed MSO, and does not reject an expired `exp`. The new test fixture demonstrates the gap: its supposedly valid Status List JWT contains only `status_list` and is accepted.
>
> Token Status List Sections 5.1 and 8.3 require `sub` and `iat`, require `sub` to equal the URI in the Referenced Token, and require `exp` to be checked when present. Without those checks, a stale signed token or a valid signed token for another list can be used to read a `VALID` bit even when the referenced current list marks this credential as revoked.
>
Please fix this once in the shared `ReferencedTokenValidator`, rather than adding mDoc-specific validation: pass the expected URI into the fetched-token validation, require correctly typed `sub` and `iat`, compare `sub` exactly with that URI, and reject expired `exp`. Update the mDoc fixture to contain valid claims and add regression tests for missing/mismatched `sub`, missing/invalid `iat`, and expired `exp`. This will strengthen both the mDoc and existing SD-JWT paths without duplicating status logic.

forkimenjeckayang: in addition
line 89
```java
JsonNode msoPayload = extractMsoPayload(token);
```
> `verifyPresentation()` has already parsed the DeviceResponse and extracted and verified its MSO, but `extractMsoPayload(token)` reparses the same token and repeats document and `IssuerAuth` extraction. Supporting this second path also required changing `MdocVerificationContext.extractDocument()` and `extractIssuerAuth()` from `private` to `public`.

This is not a correctness blocker, but it duplicates parsing and broadens the internal API. Could `verifyPresentation()` return a verification result containing the verified MSO payload, or could `MdocVerificationContext` retain or expose that verified result, so status validation consumes the same MSO instance that was authenticated?

forkimenjeckayang (commented):
a few comments @Awambeng

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
Awambeng: > Please fix this once in the shared `ReferencedTokenValidator`, rather than adding mDoc-specific validation: pass the expected URI into the fetched-token validation, require correctly typed `sub` and `iat`, compare `sub` exactly with that URI, and reject expired `exp`. Update the mDoc fixture to contain valid claims and add regression tests for missing/mismatched `sub`, missing/invalid `iat`, and expired `exp`. This will strengthen both the mDoc and existing SD-JWT paths without duplicating status logic.

Since the same validator is shared by `SD-JWT` and `mDoc` paths, this affects both formats equally. I'll suggest having a follow-up ticket to handle this. 

@forkimenjeckayang @IngridPuppet WDYT?

Awambeng: > in addition line 89
> 
> ```java
> JsonNode msoPayload = extractMsoPayload(token);
> ```
> 
> > `verifyPresentation()` has already parsed the DeviceResponse and extracted and verified its MSO, but `extractMsoPayload(token)` reparses the same token and repeats document and `IssuerAuth` extraction. Supporting this second path also required changing `MdocVerificationContext.extractDocument()` and `extractIssuerAuth()` from `private` to `public`.
> 
> This is not a correctness blocker, but it duplicates parsing and broadens the internal API. Could `verifyPresentation()` return a verification result containing the verified MSO payload, or could `MdocVerificationContext` retain or expose that verified result, so status validation consumes the same MSO instance that was authenticated?

Agreed. The `mso` payload is now stored in the context and reused instead of being extracted again.

forkimenjeckayang (dismissed):
LGTM , unless @IngridPuppet  has more

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
IngridPuppet: > Since the same validator is shared by `SD-JWT` and `mDoc` paths, this affects both formats equally. I'll suggest having a follow-up ticket to handle this.
> 
> @forkimenjeckayang @IngridPuppet WDYT?

I agree. Any heavy logic beyond supporting the specific mDoc path should be handled separately.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
```diff
@@ -81,6 +83,7 @@ public void verifyPresentation(
 
         // Verify device key binding
         CBORPairList mso = (CBORPairList) CborUtil.unwrap(issuerAuth.getPayload());
+        storeVerifiedMsoPayload(mso);
         verifyDeviceKeyBinding(document, mso, opts);
```
IngridPuppet, line 86: Nitpick comment. Since the method is named "storeVerifiedMsoPayload", it is preferable to store it after verifying the MSO, so after the call to "verifyDeviceKeyBinding".

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
```diff
@@ -380,4 +383,20 @@ private static COSESign1 extractDeviceSignature(CBORPairList document) throws Ve
...
+            this.verifiedMsoPayload = JsonSerialization.mapper.valueToTree(new CBORParser(mso.encode()).next());
+        } catch (IOException e) {
+            throw new VerificationException("Failed to serialize verified MSO", e);
+        }
+    }
+
+    /**
+     * @see <a href="https://learn.mattr.global/docs/holding/credential-claiming-guides/revocation-status-check#how-status-information-is-stored">
+     * MATTR docs — status information in the MSO</a>
+     */
```
IngridPuppet: This comment is more appropriate in MdocCredentialVerifier. It is less helpful at this point because the payload could be accessed for other things.

Also, consider adding this link from the spec too: https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-11.html#name-referenced-token-in-cose

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
Awambeng: Thanks. I created a follow-up ticket to address it: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/135

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocRevocationStatusTest.java
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    @Test
+    public void shouldExtractMsoAsJson() throws Exception {
+        String mdoc = buildDeviceResponse(opts).encodeToBase64Url();
+        var ctx = new MdocVerificationContext(mdoc);
+        ctx.verifyPresentation(opts, null, trust);
+        JsonNode mso = ctx.getVerifiedMsoPayload();
+
+        assertNotNull(mso);
```
IngridPuppet: This assertion best lives standalone or as one of the existing tests in MdocVerificationTest. The logic is not technically related to revocation.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocRevocationStatusTest.java
```diff
@@ -0,0 +1,234 @@
...
+        assertNotNull(mso.get("valueDigests"));
+        assertNotNull(mso.get("deviceKeyInfo"));
+        assertNotNull(mso.get("validityInfo"));
+    }
+
+    @Test
+    public void shouldVerifyMdoc_WhenRevocationEnforcementDisabled() throws Exception {
+        String mdoc = buildDeviceResponse(opts).encodeToBase64Url();
+        assertDoesNotThrow(() -> new MdocVerificationContext(mdoc).verifyPresentation(opts, null, trust));
+    }
```
IngridPuppet: Not sure what new this test covers. Could you clarify?

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocRevocationStatusTest.java
```diff
@@ -0,0 +1,234 @@
...
+        var ctx = new MdocVerificationContext(mdoc);
+        ctx.verifyPresentation(opts, null, trust);
+        assertDoesNotThrow(() -> new ReferencedTokenValidator(mockFetcher).validate(ctx.getVerifiedMsoPayload()));
+    }
+
+    @Test
+    public void shouldConstructVerifier_WithFetcher() throws Exception {
+        var verifier = new MdocCredentialVerifier(mockFetcher);
+        assertEquals(CredentialFormat.MSO_MDOC, verifier.format());
+    }
```
IngridPuppet: This is noise.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocRevocationStatusTest.java
```diff
@@ -0,0 +1,234 @@
...
+                        issuerSigned.findByKey(MdocConstants.L_NAME_SPACES).getValue(),
+                newIssuerAuth);
+
+        DeviceResponse newDr = new DeviceResponse(List.of(new Document(
+                DOC_TYPE,
+                newIssuerSigned,
+                (DeviceSigned) doc.findByKey(MdocConstants.L_DEVICE_SIGNED).getValue(),
+                null)));
+        return newDr.encodeToBase64Url();
+    }
```
IngridPuppet, line 152: Most of the code in this method, enabling to append identional information to the MSO payload, could live directly in MdocBaseTest. Could you consider refactoring? I bet it would be reused in the future for another test scenario.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
Awambeng: Done.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocRevocationStatusTest.java
Awambeng: Agreed. I've move the test to the `MdocVerificationTest` class.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocRevocationStatusTest.java
Awambeng: You're right. I've removed all the unnecessary tests.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocRevocationStatusTest.java
Awambeng: Thanks. Fixed.

IngridPuppet (approved):
Thank you for the update.

PR #132, opened 2026-07-21
Publish shaded releases

### LINKED ISSUE #130: Release shaded plugin to Maven
A shaded jar should be published to simplify deployments.

Related to #125 but does not aim to address the issue with `./mvnw package` broken.

ETA: 2 hours

IngridPuppet:
Closes #130

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -226,9 +226,14 @@
                             <goal>shade</goal>
                         </goals>
                         <configuration>
-                            <shadedArtifactAttached>true</shadedArtifactAttached>
-                            <shadedClassifierName>all</shadedClassifierName>
+                            <minimizeJar>true</minimizeJar>
                             <createDependencyReducedPom>false</createDependencyReducedPom>
+                            <artifactSet>
+                                <excludes>
+                                    <exclude>org.bouncycastle:*</exclude>
+                                    <exclude>com.fasterxml.jackson.*:*</exclude>
+                                </excludes>
+                            </artifactSet>
                             <filters>
                                 <filter>
                                     <artifact>*:*</artifact>
--- a/pom.xml
+++ b/pom.xml
@@ -232,6 +232,9 @@
                                 <excludes>
                                     <exclude>org.bouncycastle:*</exclude>
                                     <exclude>com.fasterxml.jackson.*:*</exclude>
+                                    <exclude>com.google.code.gson:gson</exclude>
+                                    <exclude>org.slf4j:slf4j-api</exclude>
+                                    <exclude>org.yaml:snakeyaml</exclude>
                                 </excludes>
                             </artifactSet>
                             <filters>
--- a/src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/BaseKeycloakTest.java
+++ b/src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/BaseKeycloakTest.java
@@ -46,7 +46,7 @@ public abstract class BaseKeycloakTest {
     private static final Logger logger = Logger.getLogger(BaseKeycloakTest.class);
 
     public static final String TEST_KEYCLOAK_IMAGE = "quay.io/keycloak/keycloak:26.7.0";
-    public static final String TEST_SHADED_PLUGIN_JAR = "target/keycloak-oid4vp-plugin-999.0.0-SNAPSHOT-all.jar";
+    public static final String TEST_SHADED_PLUGIN_JAR = "target/keycloak-oid4vp-plugin-999.0.0-SNAPSHOT.jar";
 
     public static final String TEST_REALM_NAME = "test";
     public static final String TEST_REALM_HAIP_NAME = "test-haip";
--- a/src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/BaseKeycloakTest.java
+++ b/src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/BaseKeycloakTest.java
@@ -69,6 +69,8 @@ public abstract class BaseKeycloakTest {
         container
                 .withImagePullPolicy(PullPolicy.alwaysPull())
                 .withProviderLibsFrom(List.of(loadShadedPluginJar()))
+                // Hot-reload main classes from "target/classes" to test implementation changes
+                // without rebuilding the jar.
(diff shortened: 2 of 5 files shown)


pom.xml
```diff
@@ -226,9 +226,14 @@
...
                         <configuration>
-                            <shadedArtifactAttached>true</shadedArtifactAttached>
-                            <shadedClassifierName>all</shadedClassifierName>
+                            <minimizeJar>true</minimizeJar>
                             <createDependencyReducedPom>false</createDependencyReducedPom>
+                            <artifactSet>
+                                <excludes>
+                                    <exclude>org.bouncycastle:*</exclude>
+                                    <exclude>com.fasterxml.jackson.*:*</exclude>
+                                </excludes>
```
forkimenjeckayang, line 238: > These exclusions do not yet produce a safe Keycloak provider artifact. Building this branch shows that the new main JAR still contains 52 `org/slf4j` entries, 158 `org/yaml` entries from SnakeYAML, and 11 minimized `com/google` entries from Gson. The target Keycloak 26.7.0 image already contains `slf4j-api-2.0.17`, SnakeYAML 2.0, and Gson 2.13.2, while this JAR embeds Gson 2.10.1. Keycloak's provider documentation says provider JARs share the application classloader, are given precedence over built-in libraries, and must not include classes or resources that conflict with server libraries. A partial older Gson package is particularly risky because it can create a mixed-version package at runtime; the embedded SLF4J and SnakeYAML classes are duplicate server classes as well.

Could we constrain the shade to dependencies Keycloak does not provide, or relocate the private bundled libraries. At minimum, exclude the Keycloak-provided Gson, SLF4J, and SnakeYAML artifacts and add an artifact-content assertion that fails if known server packages such as `org/slf4j`, `org/yaml/snakeyaml`, `com/google/gson`, `com/fasterxml/jackson`, or `org/bouncycastle` appear in the release JAR.

pom.xml
```diff
@@ -226,9 +226,14 @@
                             <goal>shade</goal>
                         </goals>
                         <configuration>
-                            <shadedArtifactAttached>true</shadedArtifactAttached>
-                            <shadedClassifierName>all</shadedClassifierName>
+                            <minimizeJar>true</minimizeJar>
                             <createDependencyReducedPom>false</createDependencyReducedPom>
```
forkimenjeckayang, line 230: > The shaded JAR now replaces the project's main artifact, but `createDependencyReducedPom=false` leaves `com.authlete:cbor` and `com.networknt:json-schema-validator` as normal compile dependencies in the published flattened POM. Both libraries are already embedded in the 1.3 MB main JAR. A Maven consumer will therefore resolve standalone copies of those dependencies and their transitives alongside the same unrelocated classes inside the plugin JAR, defeating the single-artifact release and creating duplicate class definitions.

Could we publish dependency metadata that reflects the actual artifact: generate and deploy a dependency-reduced POM, or otherwise mark/remove dependencies that are bundled while retaining only genuinely external/provided requirements. Add an install-stage test that inspects the installed POM and confirms bundled dependencies are not exposed transitively.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/BaseKeycloakTest.java
```diff
@@ -46,7 +46,7 @@ public abstract class BaseKeycloakTest {
     private static final Logger logger = Logger.getLogger(BaseKeycloakTest.class);
 
     public static final String TEST_KEYCLOAK_IMAGE = "quay.io/keycloak/keycloak:26.7.0";
-    public static final String TEST_SHADED_PLUGIN_JAR = "target/keycloak-oid4vp-plugin-999.0.0-SNAPSHOT-all.jar";
+    public static final String TEST_SHADED_PLUGIN_JAR = "target/keycloak-oid4vp-plugin-999.0.0-SNAPSHOT.jar";
```
forkimenjeckayang, line 49: > Pointing the harness at the new filename is useful, but this container still calls `withProviderClassesFrom("target/classes", "target/test-classes")`. Consequently, all production plugin classes are injected into separate provider JARs in addition to the shaded artifact. The Keycloak startup log confirms split packages across `0_providers.jar`, `1_providers.jar`, and the release JAR, so this suite does not prove that the published artifact alone contains and discovers every production provider correctly.

Could we keep the existing fixture setup where the test authenticator is required, but add a small release-artifact smoke test that starts Keycloak with only `keycloak-oid4vp-plugin-${version}.jar` mounted and verifies the production provider factories are discovered?  That test should also be the place to enforce the forbidden-package checks above.

forkimenjeckayang (commented):
a few comments from my agent 😄  after some analysis 
WDYT @IngridPuppet ?

pom.xml
IngridPuppet: I'd rather not reduce the pom to for dependency security checks.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/BaseKeycloakTest.java
IngridPuppet: The plugin jar is mounted only for transitive dependencies to be loaded. It is still useful to keep `withProviderClassesFrom("target/classes", "target/test-classes")` for hot testing after an implementation change that would require to compile a new jar. I'll comment that. Testing whether the jar alone is functional is overkill... We've had no such test since the beginning. An issue where a given business class is not included in the jar is unlikely to occur.

pom.xml
IngridPuppet: > Could we constrain the shade to dependencies Keycloak does not provide?

That's exactly what we're trying to do. The problem is that transitive deps are more difficult to process via includes rather than excludes. And missing excludes are less problematic than missing includes with upgrades.

> At minimum, exclude the Keycloak-provided Gson, SLF4J, and SnakeYAML artifacts.

Done. My agent just missed them :sweat_smile: There are also risks with exclusions. Hopefully, our integration tests constitute a safe net.

> add an artifact-content assertion that fails if known server packages

This is overkill. The shade plugin directives are well trusted. We're safe just paying attention to the final jar size, which about 1MB currently. That's good.

forkimenjeckayang (approved):
LGTM

IngridPuppet:
Thank you for the review.

PR #137, opened 2026-07-23
Reduce reliance on session notes

### LINKED ISSUE #122: Reduce reliance on session notes to pass data to authenticator
Follow up to:
- #88 

These may be retrieved directly from the authorization context for example.

```java
        DcqlQuery dcqlQuery = authContext.getRequestObject().getDcqlQuery();
        boolean requireCryptographicHolderBinding = dcqlQuery.getCredentials().stream()
                .noneMatch(c -> Boolean.FALSE.equals(c.getRequireCryptographicHolderBinding()));
        authSession.setAuthNote(
                OID4VPAuthenticator.REQUIRE_CRYPTOGRAPHIC_HOLDER_BINDING_KEY,
                String.valueOf(requireCryptographicHolderBinding));

        var transactionData = authContext.getRequestObject().getTransactionData();
        if (transactionData != null && !transactionData.isEmpty()) {
            try {
                authSession.setAuthNote(
                        OID4VPAuthenticator.TRANSACTION_DATA_WIRE_KEY,
                        JsonSerialization.writeValueAsString(transactionData));
            } catch (Exception e) {
                throw new IllegalStateException("Failed to persist transaction_data for validation", e);
            }
        }
```

ETA: 2 hours

IngridPuppet:
Closes #122

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonNode;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.binding.BindingValueComparator;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.binding.ExactBindingValueComparatorFactory;
+import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.config.AuthRequirements;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model.dcql.Credential;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model.dcql.DcqlQuery;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model.dto.AuthorizationContext;
@@ -55,10 +56,6 @@ public class OID4VPAuthenticator implements Authenticator {
      */
     public static final String PRESENTED_TOKENS_KEY = "presented_tokens";
 
-    public static final String REQUIRE_CRYPTOGRAPHIC_HOLDER_BINDING_KEY = "require_cryptographic_holder_binding";
-
-    public static final String TRANSACTION_DATA_WIRE_KEY = "transaction_data_wire";
-
     public OID4VPAuthenticator(Map<String, CredentialVerifier> handlers) {
         this.handlers = handlers;
     }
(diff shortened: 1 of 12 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
```diff
@@ -71,7 +68,7 @@ public void authenticate(AuthenticationFlowContext context) {
         CredentialRequirement primaryCredential = profile.getPrimaryCredential();
 
         AuthorizationContext authContext = getAuthorizationContext(authSession);
-        boolean requireCryptographicHolderBinding = parseRequireCryptographicHolderBinding(authSession);
+        AuthRequirements authRequirements = new AuthRequirements(context.getAuthenticatorConfig());
```
forkimenjeckayang, line 72: > This removes the holder-binding session note, but replaces its request-derived value with the current authenticator configuration. `require_cryptographic_holder_binding` is a property of each DCQL Credential Query (default `true`), and the persisted `AuthorizationContext.requestObject` is the immutable request the Wallet answered. The realm configuration can change while that request is in flight.
>
Could we  derive the holder-binding requirement from the matching credential query in `authorizationContext.getRequestObject().getDcqlQuery()` using `credential.getId()`, defaulting to `true` when the field is absent. Realm configuration should remain the source used to construct new requests, while verification should use the stored request. 

Also adding regression tests where the stored query and current config disagree in both directions, and cover credential-specific values in a multi-credential query ?

forkimenjeckayang (commented):
one concern @IngridPuppet 
Could you please check?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
IngridPuppet: The current PR precisely avoids treating the DCQL query generated as the source of truth for anything. It is just a mediator. The ground truth remains the configuration. Currently, claim requirement enforcement, trust enforcement, and several other checks are governed by the config, which is the expected behavior.

An authentication session is short-lived, so mismatch between the DCQL query and the config is an edge case. Should it occur, the config should still prevail because it was changed for a reason.

Lastly, there is already such defensive check (at least for SD-JWT) considering the requirement in the DCQL.

https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/blob/1bc8a3e83c3b2205d34e7683f4b1d02d1ba4ca9f/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/DcqlPresentationValidator.java#L48-L54

forkimenjeckayang (dismissed):
LGTM

forkimenjeckayang (approved):

PR #138, opened 2026-07-23
Implement transaction_data binding verification for mso_mdoc credentials

### LINKED ISSUE #124: Implement transaction_data binding verification for mso_mdoc credentials
Follow up to:
- #88 

Self-explanatory. Covered for SD-JWT but not mDoc.

ETA: 1 day.

Awambeng:
This PR adds `transaction_data_hashes` validation for `mso_mdoc` credentials, analogous to the existing SD-JWT KB-JWT validation.

The wallet includes the hashes in `DeviceSigned.nameSpaces`, where they are covered by the device signature. During presentation verification, the verifier extracts these hashes and validates them against the transaction data supplied in the
authorization request.

### Changes

- **VerifierConfig**: Removed the guard preventing `transaction_data` from being used with `mso_mdoc` as the primary credential.
- **MdocVerificationContext**: Added `getDeviceNameSpaces()` to expose `DeviceSigned.nameSpaces` during verification.
- **MdocCredentialVerifier**: Added `validateTransactionData()`, which reads `transaction_data_hashes` from the device namespaces and delegates validation to `TransactionDataValidator`.
- **MdocBaseTest**: Added builder support for custom `DeviceNameSpaces` in test mDocs.
- **Tests**: Updated the profile configuration test and added four transaction data verification tests for mDoc credentials.

Closes #124

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
@@ -50,6 +50,7 @@ public class MdocVerificationContext {
 
     private final CBORPairList mdoc;
     private JsonNode verifiedMsoPayload;
+    private CBORTaggedItem deviceNameSpaces;
 
     public MdocVerificationContext(String mdoc) throws VerificationException {
         try {
@@ -57,16 +58,6 @@ public class MdocVerificationContext {
         } catch (MdocEncodingException e) {
             throw new VerificationException("Failed to parse subject as an mDoc device response", e);
         }
-    }
-
-    /**
-     * Returns the DeviceSigned namespaces from the parsed mDoc for use during
-     * presentation verification, including inspection of elements such as
-     * {@code transaction_data_hashes}.
-     */
-    public CBORTaggedItem getDeviceNameSpaces() throws VerificationException {
-        CBORPairList document = extractDocument(mdoc);
-        return extractDeviceNamespaces(document);
     }
 
     /**
@@ -90,6 +81,9 @@ public class MdocVerificationContext {
         CBORPairList document = extractDocument(mdoc);
         COSESign1 issuerAuth = extractIssuerAuth(document);
         verifyIssuerSignature(issuerAuth, truststoreProvider);
+
+        // Cache for getDeviceNameSpaces() to avoid re-parsing
+        this.deviceNameSpaces = extractDeviceNamespaces(document);
 
         // Verify device key binding
         CBORPairList mso = (CBORPairList) CborUtil.unwrap(issuerAuth.getPayload());
@@ -402,7 +396,18 @@ public class MdocVerificationContext {
         }
     }
 
+    /**
+     * Returns the verified MSO payload as JSON.
+     */
     public JsonNode getVerifiedMsoPayload() {
         return verifiedMsoPayload;
+    }
+
+    /**
+     * Returns the DeviceSigned nameSpaces from the parsed mDoc, cached during
+     * {@link #verifyPresentation(MdocVerificationOpts, PresentationRequirements, TrustAnchorProvider)}.
+     */
+    public CBORTaggedItem getDeviceNameSpaces() {
+        return deviceNameSpaces;
     }
 }
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
@@ -60,6 +60,16 @@ public class MdocVerificationContext {
     }
 
     /**
+     * Returns the DeviceSigned namespaces from the parsed mDoc for use during
+     * presentation verification, including inspection of elements such as
+     * {@code transaction_data_hashes}.
+     */
+    public CBORTaggedItem getDeviceNameSpaces() throws VerificationException {
+        CBORPairList document = extractDocument(mdoc);
+        return extractDeviceNamespaces(document);
+    }
+
+    /**
      * Verifies mDoc presentation.
      *
      * @param opts                            Options to parameterize the mDoc verification.
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
@@ -86,9 +86,6 @@ public class OID4VPAuthenticator implements Authenticator {
         try {
             primaryClaims = primaryVerifier.verifyCredential(
                     context, authContext, authRequirements, primaryCredential, primaryToken);
-            // SD-JWT validates transaction_data via the interface method (above).
-            // mso_mdoc validates internally inside verifyCredential() because it
-            // depends on the post-verification MdocVerificationContext state.
             primaryVerifier.validateTransactionData(authContext, primaryToken);
         } catch (VerificationException e) {
             logger.errorf(e, "Primary credential verification failed (authSession = %s)", correlationId(context));
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
@@ -86,6 +86,7 @@ public class OID4VPAuthenticator implements Authenticator {
         try {
             primaryClaims = primaryVerifier.verifyCredential(
                     context, authContext, authRequirements, primaryCredential, primaryToken);
+            primaryVerifier.validateTransactionData(authContext, primaryToken);
         } catch (VerificationException e) {
             logger.errorf(e, "Primary credential verification failed (authSession = %s)", correlationId(context));
             failRejectingPresentedCredential(context, e.getMessage(), e);
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
@@ -86,6 +86,9 @@ public class OID4VPAuthenticator implements Authenticator {
         try {
             primaryClaims = primaryVerifier.verifyCredential(
                     context, authContext, authRequirements, primaryCredential, primaryToken);
+            // SD-JWT validates transaction_data via the interface method (above).
+            // mso_mdoc validates internally inside verifyCredential() because it
+            // depends on the post-verification MdocVerificationContext state.
             primaryVerifier.validateTransactionData(authContext, primaryToken);
         } catch (VerificationException e) {
             logger.errorf(e, "Primary credential verification failed (authSession = %s)", correlationId(context));
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
@@ -2,7 +2,17 @@ package io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.authenticator.mdoc;
 
 import static io.github.adorsysgis.keycloak.protocol.oid4vc.mdoc.MdocConstants.L_NAME_SPACES;
 
+import com.authlete.cbor.CBORItem;
+import com.authlete.cbor.CBORItemList;
+import com.authlete.cbor.CBORPair;
+import com.authlete.cbor.CBORPairList;
+import com.authlete.cbor.CBORString;
+import com.authlete.cbor.CBORTaggedItem;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.github.adorsysgis.keycloak.protocol.oid4vc.mdoc.CborUtil;
+import io.github.adorsysgis.keycloak.protocol.oid4vc.mdoc.MdocConstants;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.mdoc.MdocVerificationContext;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.mdoc.MdocVerificationOpts;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.authenticator.CredentialFormat;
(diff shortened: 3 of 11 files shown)


Awambeng:
> ## Code Review Summary
> ### Critical Issue
> **Algorithm Confusion Attack** (`MdocCredentialVerifier.java:178-181`): The hash algorithm is extracted from the client-supplied mDoc and passed to `TransactionDataValidator`. An attacker could use a weaker algorithm (e.g., MD5) to circumvent transaction data binding security. The verifier should dictate the expected algorithm, not trust the client's `TRANSACTION_DATA_HASHES_ALG_CLAIM`.
> 
The hash algorithm is validated against both the verifier's request-specified `transaction_data_hashes_alg` (which dictates the allowed set) and `TransactionDataSupport.validateHashAlgorithm()`  which only accepts  sha-256  (hardcoded). The wallet's claimed algorithm must match the verifier's allowed algorithms. No attack vector exists with the current supported algorithms.

> ### Medium Issues
> 1. **Silent Hash Filtering** (`extractHashesArray()`): Non-`CBORString` items are dropped without warning or error. Malformed data silently reduces validation scope.
>
Fixed. `VerificationException` is thrown when encountering non-string entries in the hashes array, preventing silent hash filtering.

> 2. **Namespace Loop Overwrites** (`validateTransactionData()`): If `transaction_data_hashes` appears in multiple namespaces, only the last one's hashes are used. This could allow an attacker to hide hashes in multiple namespaces.
> 
Fixed. We now reject transaction data hashes appearing in multiple namespaces with a descriptive error.

> ### Missing Test Coverage
> * Algorithm mismatch scenarios
> * Malformed CBOR hash structures
> * End-to-end `verifyCredential()` flow with both revocation and transaction data
> 
Fixed. Added test cases for:
- Malformed CBOR hashes
- Duplicate namespaces
- End-to-end test for revocation and transaction data

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
```diff
@@ -95,9 +112,110 @@ public JsonNode verifyCredential(
...
+        }
+
+        ObjectNode hashesPayload = extractTransactionDataHashes(verificationContext);
+        if (hashesPayload == null) {
+            throw new VerificationException(
+                    "Device namespaces must contain transaction_data_hashes when transaction_data is requested");
+        }
+
+        try {
+            TransactionDataValidator.validate(transactionDataWire, hashesPayload);
```
forkimenjeckayang, line 148: > This proves that the hashes are covered by DeviceAuth, but it does not prove that the issuer authorized the device key to sign this transaction-data element. Final 1.0 Appendix B.2.1 defines mDoc transaction-data binding through a transaction-type-specific `(NameSpace, DataElementIdentifier, DataElementValue)` and says the issuer includes the relevant element in `DeviceKeyInfo.keyAuthorizations`. The verifier currently scans for the SD-JWT name `transaction_data_hashes` in any namespace and never inspects `keyAuthorizations` or maps the requested `type` to an mDoc transaction-data profile.
>
> The positive fixture demonstrates the gap: `MdocBaseTest` authorizes only the issuer-claim namespace `com.example.namespace1`, while `buildMdocWithTransactionData()` places the hashes under `com.example.doctype`; the value is accepted despite that namespace not being issuer-authorized.
>
Please model the supported mDoc transaction-data profile(s), mapping each supported request `type` to its exact namespace, element identifier, value structure, and processing/hash rules. Before accepting the binding, require that exact element or its namespace is present in the issuer-signed MSO `KeyAuthorizations`; reject unconfigured types or unauthorized elements. Add full-verifier tests for authorized, unauthorized, missing, and mismatched elements. Reusing the shared hash primitive is fine once the mDoc profile has selected and authorized the value.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
```diff
@@ -95,9 +112,110 @@ public JsonNode verifyCredential(
...
+            if (hashesPair != null) {
+                if (hashes != null) {
+                    throw new VerificationException("transaction_data_hashes must not appear in multiple namespaces");
+                }
+                hashes = extractHashesArray(hashesPair.getValue());
+            }
+            var algPair = itemsMap.findByKey(TransactionDataValidator.TRANSACTION_DATA_HASHES_ALG_CLAIM);
+            if (algPair != null && algPair.getValue() instanceof CBORString algStr) {
+                hashesAlg = algStr.getValue();
+            }
```
forkimenjeckayang, line 196: > This loop loses the namespace association required by Appendix B.2.1. It can take `transaction_data_hashes` from namespace A and `transaction_data_hashes_alg` from namespace B, then synthesize one payload that never existed as a single mDoc data-element tuple. Multiple algorithm entries are also silently resolved by taking the last one.
>
 Please resolve one exact profile namespace first and read all related elements only from that namespace. Reject duplicate, cross-namespace, wrong-type, or unexpected algorithm elements. Add regressions where the hash and algorithm are split across namespaces and where multiple namespaces supply an algorithm.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocRevocationStatusTest.java
```diff
@@ -141,6 +150,77 @@ public void shouldFail_WhenRevocationEnforcedAndStatusInvalid() throws Exception
...
+                .thenReturn(JsonSerialization.writeValueAsString(List.of(wire)));
+
+        var context = mock(AuthenticationFlowContext.class);
+        when(context.getAuthenticatorConfig()).thenReturn(authConfig);
+        when(context.getAuthenticationSession()).thenReturn(authSession);
+
+        var credential = new CredentialRequirement()
+                .setId("test")
+                .setCredentialTypes(List.of(DOC_TYPE))
+                .setTrust(List.of(new TrustPolicy().setType(TrustPolicy.X5C).setAnchors(List.of(getIssuerCertRef1()))));
```
forkimenjeckayang, line 190: > This credential keeps the default `SUPPORTING` role, but production calls `validateTransactionData()` only when `credential.isPrimary()` is true. Consequently, `shouldVerifyCredential_WithBothRevocationAndTransactionData` verifies revocation only; removing or breaking the new transaction-data call would not fail this test. The other new tests invoke the package-private helper directly without first verifying issuer signature, DeviceAuth, or the MSO.
>
 Please set this requirement to `CredentialRole.PRIMARY` and assert that a mismatched transaction hash is rejected through `verifyCredential()`. The positive full-flow fixture should also carry a KeyAuthorization for the selected transaction-data element, with a negative fixture proving that an otherwise valid DeviceSignature is rejected when that authorization is absent.

forkimenjeckayang (commented):
a few comments

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocRevocationStatusTest.java
Awambeng: Thanks. Changes applied as suggested.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
Awambeng: Refactored to a two-pass approach: first resolves the single namespace containing  transaction_data_hashes , then reads  transaction_data_hashes_alg  only from that namespace. Rejects alg in any other namespace, rejects non-string alg values, and handles both bare CBOR string and single-element array encodings (authlete  DeviceSignedItemsEntry  wrapping) via  extractAlgString() . Regression test added for split alg/hashes namespaces.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
Awambeng: Added `verifyNamespaceIsAuthorized()` checks that the namespace containing `transaction_data_hashes` is present in the MSO's `deviceKeyInfo.keyAuthorizations.nameSpaces` before accepting the binding. This addresses the issuer authorization gap with positive (authorized namespace) and negative (unauthorized namespace) verifier tests. Broader transaction-data profile mapping can be introduced separately if needed.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
```diff
@@ -95,9 +113,168 @@ public JsonNode verifyCredential(
...
+        for (JsonNode ns : authorizedNs) {
+            if (ns.asText().equals(namespace)) {
+                authorized = true;
+                break;
+            }
+        }
+        if (!authorized) {
+            throw new VerificationException(
+                    "Namespace '" + namespace + "' is not authorized for transaction_data_hashes in the MSO");
+        }
```
forkimenjeckayang: > This now checks MSO `KeyAuthorizations`, but only the whole-namespace form. ISO mdoc also permits `keyAuthorizations.dataElements`, where the device key is authorized for named elements without being authorized for the entire namespace. A credential whose MSO authorizes `transaction_data_hashes` exactly through `dataElements` is therefore rejected as missing authorized `nameSpaces`.

Please accept an element when either its namespace is listed in `nameSpaces` or the element identifier is listed under `dataElements[namespace]`. Apply that check to every DeviceSigned element consumed here: `transaction_data_hashes`, and `transaction_data_hashes_alg` when it is present. Add a positive `dataElements`-only test and negative tests where the hash or algorithm element is not authorized.

forkimenjeckayang (commented):
one last comment

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
Awambeng: fixed.

forkimenjeckayang (dismissed):
Thanks for addressing my concerns

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocVerificationContext.java
```diff
@@ -59,6 +59,16 @@ public MdocVerificationContext(String mdoc) throws VerificationException {
...
 
+    /**
+     * Returns the DeviceSigned namespaces from the parsed mDoc for use during
+     * presentation verification, including inspection of elements such as
+     * {@code transaction_data_hashes}.
+     */
+    public CBORTaggedItem getDeviceNameSpaces() throws VerificationException {
+        CBORPairList document = extractDocument(mdoc);
+        return extractDeviceNamespaces(document);
+    }
```
IngridPuppet: Could you relocate this getter next to `getVerifiedMsoPayload`? It helps readability to keep similar methods close. Also, this parses the mdoc tree once again. Could you check if it's possible to use the same pattern with the MSO payload?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
```diff
@@ -96,9 +109,198 @@ public JsonNode verifyCredential(
             }
         }
 
+        if (credential.isPrimary()) {
+            validateTransactionData(authorizationContext, verificationContext);
+        }
```
IngridPuppet: Because both verifiers do this, what about adding `validateTransactionData` to the contract of a CredentialVerifier and call this directly in OID4VPAuthenticator? So we're sure future formats do not omit to make this call.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
```diff
@@ -96,9 +109,198 @@ public JsonNode verifyCredential(
...
+            var algPair = itemsMap.findByKey(TransactionDataValidator.TRANSACTION_DATA_HASHES_ALG_CLAIM);
+            if (algPair == null) {
+                continue;
+            }
+            if (!nsName.equals(hashesNamespace)) {
+                throw new VerificationException(
+                        "transaction_data_hashes_alg must not appear outside the namespace containing transaction_data_hashes");
+            }
+            hashesAlg = extractAlgString(algPair.getValue());
+        }
```
IngridPuppet: Could you optimize these passes into just one? For each potential namespace, you collect both values while checking that the current namespace is equal to hashesNamespace if not null. Or a similar logic.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
```diff
@@ -96,9 +109,198 @@ public JsonNode verifyCredential(
...
+        if (dki == null) {
+            throw new VerificationException("MSO is missing deviceKeyInfo");
+        }
+        JsonNode keyAuth = dki.get("keyAuthorizations");
+        if (keyAuth == null) {
+            throw new VerificationException("MSO deviceKeyInfo is missing keyAuthorizations");
+        }
+        // Check namespace-level authorization first
+        JsonNode nameSpaces = keyAuth.get("nameSpaces");
+        if (nameSpaces != null && nameSpaces.isArray()) {
```
IngridPuppet, line 258: Use constants from MdocConstants. If not available, add them.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/mdoc/MdocCredentialVerifier.java
```diff
@@ -96,9 +109,198 @@ public JsonNode verifyCredential(
...
+                    }
+                    if (!found) {
+                        allFound = false;
+                        break;
+                    }
+                }
+                if (allFound) {
+                    return;
+                }
+            }
```
IngridPuppet, line 276: These nested loops are not very friendly nor efficient. Could you explore leveraging the Java Stream API to achieve the same?

```java
if (elements != null && elements.isArray()) {
    Set<String> existingIds = StreamSupport.stream(elements.spliterator(), false)
            .map(JsonNode::asText)
            .collect(Collectors.toSet());

    if (existingIds.containsAll(elementIds)) {
        return;
    }
}
```

IngridPuppet (changes requested):

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
```diff
@@ -86,6 +86,10 @@ public void authenticate(AuthenticationFlowContext context) {
         try {
             primaryClaims = primaryVerifier.verifyCredential(
                     context, authContext, authRequirements, primaryCredential, primaryToken);
+            // SD-JWT validates transaction_data via the interface method (above).
+            // mso_mdoc validates internally inside verifyCredential() because it
+            // depends on the post-verification MdocVerificationContext state.
+            primaryVerifier.validateTransactionData(authContext, primaryToken);
```
IngridPuppet: This behavior is strange. All implementations must behave the same. Can't MdocCredentialVerifier just keep the post verification context as a property in the verifier? Also there should be no default implementation of `validateTransactionData` in the interface, so a no-op impl must be explicit not implicit.

IngridPuppet (commented):
Just one last architectural comment...

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
Awambeng: Addressed.

- Removed the default validateTransactionData implementation. It is now required for all implementations.
- MdocCredentialVerifier now stores the MdocVerificationContext during verifyCredential() and uses it for transaction validation. Removed the credential.isPrimary() check.

IngridPuppet (approved):
Thank you very much for the update.

PR #141, opened 2026-07-28
Implement migration strategy for changes to authenticator configuration

### LINKED ISSUE #139: Implement migration strategy for the authenticator setup and configuration
There are more broken realms than foreseen. We need such migration. Focus: 1.2.6 to 1.3.0.

```
[org.keycloak.quarkus.runtime.cli.ExecutionExceptionHandler] (executor-thread-1) ERROR: Authentication flow 'oid4vp auth' already exists in realm 'master' but does not contain the expected OpenID4VP authenticator 'oid4vp-authenticator'. Refusing to start. Consider backing up the current flow configuration and removing it so Keycloak can create a new one on next restart.
```

ETA: 1 day

IngridPuppet:
Closes #139

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointFactory.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointFactory.java
@@ -1,17 +1,12 @@
 package io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp;
 
-import static io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.OID4VPUserAuthEndpointBase.OID4VP_AUTH_FLOW;
-
 import io.github.adorsysgis.keycloak.protocol.oid4vc.crypto.ExtendedCertificateUtils;
-import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.authenticator.OID4VPAuthenticatorFactory;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.config.OID4VPConfig;
+import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.migration.OID4VPMigrationManager;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.utils.ErrorResponseSanitizer;
 import org.jboss.logging.Logger;
 import org.keycloak.Config;
-import org.keycloak.authentication.AuthenticationFlow;
 import org.keycloak.events.EventBuilder;
-import org.keycloak.models.AuthenticationExecutionModel;
-import org.keycloak.models.AuthenticationFlowModel;
 import org.keycloak.models.KeycloakContext;
 import org.keycloak.models.KeycloakSession;
 import org.keycloak.models.KeycloakSessionFactory;
@@ -32,6 +27,8 @@ public class OID4VPUserAuthEndpointFactory implements RealmResourceProviderFacto
 
     public static final String PROVIDER_ID = "oid4vp-auth";
 
+    private OID4VPConfig pluginConfig;
+
     @Override
     public String getId() {
         return PROVIDER_ID;
@@ -47,68 +44,27 @@ public class OID4VPUserAuthEndpointFactory implements RealmResourceProviderFacto
 
     @Override
     public void init(Config.Scope config) {
-        OID4VPConfig pluginConfig = new OID4VPConfig(config);
-        ExtendedCertificateUtils.init(pluginConfig);
-        ErrorResponseSanitizer.init(pluginConfig);
+        this.pluginConfig = new OID4VPConfig(config);
+        ExtendedCertificateUtils.init(this.pluginConfig);
+        ErrorResponseSanitizer.init(this.pluginConfig);
     }
 
     @Override
     public void postInit(KeycloakSessionFactory factory) {
+        OID4VPMigrationManager migrationManager = new OID4VPMigrationManager(pluginConfig);
+
         factory.register(event -> {
             if (event instanceof PostMigrationEvent) {
                 logger.debugf("Migrating existing realms to add OpenID4VP user auth flow...");
-                KeycloakModelUtils.runJobInTransaction(
-                        factory, session -> session.realms().getRealmsStream().forEach(this::migrateRealmIfNecessary));
+                KeycloakModelUtils.runJobInTransaction(factory, session -> session.realms()
+                        .getRealmsStream()
+                        .forEach(realm -> migrationManager.migrate(session, realm)));
             } else if (event instanceof RealmModel.RealmPostCreateEvent realmEvent) {
                 logger.debugf("Migrating newly created realm to add OpenID4VP user auth flow...");
                 RealmModel realm = realmEvent.getCreatedRealm();
-                migrateRealmIfNecessary(realm);
+                migrationManager.migrate(realmEvent.getKeycloakSession(), realm);
             }
         });
-    }
-
-    private void migrateRealmIfNecessary(RealmModel realm) {
-        AuthenticationFlowModel oid4vpAuthFlow = realm.getFlowByAlias(OID4VP_AUTH_FLOW);
-        if (oid4vpAuthFlow == null) {
-            logger.infof("Creating default OpenID4VP user auth flow for realm '%s'", realm.getName());
-            oid4vpAuthenticationFlow(realm);
-            return;
-        }
-
-        logger.debugf("OpenID4VP user auth flow already exists for realm '%s'", realm.getName());
-
-        boolean hasExpectedAuthenticator = realm.getAuthenticationExecutionsStream(oid4vpAuthFlow.getId())
-                .map(AuthenticationExecutionModel::getAuthenticator)
-                .anyMatch(OID4VPAuthenticatorFactory.PROVIDER_ID::equals);
-
-        if (!hasExpectedAuthenticator) {
-            throw new IllegalStateException(String.format(
-                    "Authentication flow '%s' already exists in realm '%s' but does not contain the expected "
-                            + "OpenID4VP authenticator '%s'. Refusing to start. Consider backing up the current "
-                            + "flow configuration and removing it so Keycloak can create a new one on next restart.",
-                    OID4VP_AUTH_FLOW, realm.getName(), OID4VPAuthenticatorFactory.PROVIDER_ID));
-        }
-    }
-
-    private void oid4vpAuthenticationFlow(final RealmModel realm) {
-        AuthenticationFlowModel oid4vpAuthFlow = new AuthenticationFlowModel();
-
-        oid4vpAuthFlow.setAlias(OID4VP_AUTH_FLOW);
-        oid4vpAuthFlow.setDescription("Authenticate via OpenID4VP presentations of self-issued identity credentials");
-        oid4vpAuthFlow.setProviderId(AuthenticationFlow.BASIC_FLOW);
-        oid4vpAuthFlow.setTopLevel(true);
-        oid4vpAuthFlow.setBuiltIn(false);
-        oid4vpAuthFlow = realm.addAuthenticationFlow(oid4vpAuthFlow);
-
-        AuthenticationExecutionModel execution = new AuthenticationExecutionModel();
-
-        execution.setParentFlow(oid4vpAuthFlow.getId());
-        execution.setRequirement(AuthenticationExecutionModel.Requirement.REQUIRED);
-        execution.setAuthenticator(OID4VPAuthenticatorFactory.PROVIDER_ID);
-        execution.setPriority(10);
-        execution.setAuthenticatorFlow(false);
-
-        realm.addAuthenticatorExecution(execution);
     }
 
     @Override
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/OID4VPConfig.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/OID4VPConfig.java
@@ -2,7 +2,10 @@ package io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.config;
 
 import io.github.adorsysgis.keycloak.protocol.oid4vc.crypto.ExtendedCertificateUtils;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.OID4VPUserAuthEndpointFactory;
+import java.util.Arrays;
+import java.util.List;
 import org.keycloak.Config;
+import org.keycloak.utils.StringUtil;
 
 /**
  * Centralized configuration for this plugin, initialized from the {@code oid4vp-auth} provider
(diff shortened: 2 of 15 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/OID4VPMigrationManager.java
```diff
@@ -0,0 +1,154 @@
...
+            } catch (RuntimeException e) {
+                logger.errorf(
+                        e,
+                        "Migration '%s' failed for realm '%s'; aborting remaining migrations",
+                        migration.id(),
+                        realm.getName());
+                throw e;
+            }
+            realm.setAttribute(LAST_APPLIED_MIGRATION_ATTRIBUTE, migration.id());
+            logger.infof("Migration '%s' completed for realm '%s'", migration.id(), realm.getName());
```
forkimenjeckayang, line 140: > The legacy `sd-jwt-authenticator` case is migrated correctly, but the manager no longer enforces the invariant that the reserved `oid4vp auth` flow contains `oid4vp-authenticator`. If a realm has this flow alias with an unrelated, misspelled, or otherwise unsupported provider, `Migration_v1_3_0` performs a no-op and this loop still stamps `v1.3.0` as successfully applied. Every later restart then skips the migration, while `OID4VPUserAuthEndpointBase` falls back to an empty authenticator config and the authentication processor executes an incompatible flow.
>

`origin/main` rejected this state, while this branch now allows startup to continue and records the migration as applied. Is that relaxation intentional? The ticket's supported legacy provider is migrated correctly, so I do not consider this a blocker for the `1.2.6 -> 1.3.0` repair. As follow-up hardening, it would be useful to define the policy explicitly and add tests for legacy, current, unknown, and mixed execution sets. If unknown flows are unsupported, report them clearly and avoid presenting the realm as healthy; if custom flows are intentionally allowed, document that contract.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/OID4VPConfig.java
```diff
@@ -2,21 +2,60 @@
...
     private final boolean verboseErrors;
     private final int certificateCacheMaxSize;
+    private final List<String> autoCreateAuthFlowRealms;
+
+    public OID4VPConfig(Config.Scope config) {
+        this.autoCreateAuthFlowRealms = config == null ? List.of() : parseRealmList(config.get(MANAGED_REALMS_CONFIG));
+        this.verboseErrors = config != null && config.getBoolean(VERBOSE_ERRORS_CONFIG, false);
+        this.certificateCacheMaxSize = config == null
+                ? ExtendedCertificateUtils.DEFAULT_MAX_CACHE_SIZE
+                : config.getInt(CACHE_MAX_SIZE_CONFIG, ExtendedCertificateUtils.DEFAULT_MAX_CACHE_SIZE);
```
forkimenjeckayang, line 44: > This deliberately changes the default outside the direct `1.2.6 -> 1.3.0` migration. `origin/main` creates the `oid4vp auth` flow for every existing and newly created realm; this branch defaults `managed-realms` to an empty list. Existing legacy realms still migrate because they already contain the flow, but a realm created after upgrade no longer receives it unless operators add the new SPI option.
>

 Is that compatibility change intentional for this release? The README explains how to opt in, and the unit tests encode the empty default, so this may be a product decision rather than an implementation mistake. If intentional, please call it out explicitly in release/upgrade documentation and update deployment configurations that expect automatic provisioning. If not, preserve the previous default and provide an explicit opt-out. In either case, retain an integration test for realm creation under the chosen default.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/OID4VPMigrationManager.java
```diff
@@ -0,0 +1,154 @@
...
+        if (resumeIndex > 0) {
+            logger.debugf(
+                    "Skipping %d already-applied migration(s) for realm '%s' (lastApplied=%s)",
+                    resumeIndex, realm.getName(), lastApplied);
+        } else if (lastApplied != null) {
+            logger.warnf(
+                    "Realm '%s' has lastApplied='%s' which does not match any registered migration; "
+                            + "running all migrations from the start",
+                    realm.getName(), lastApplied);
+        }
```
forkimenjeckayang, line 124: > Treating every unknown marker as older is unsafe. If a realm has already been migrated by a newer plugin, for example `v1.4.0`, and a `v1.3.0` node starts during a rolling deployment or rollback, this code reruns all migrations and then overwrites the marker with `v1.3.0`. The next newer node reruns its migration, and mixed versions can keep moving the marker backward and forward.
>

The migration registry should be append-only, so a legitimate older applied ID should still be present and discoverable. Please fail closed or leave the realm untouched when a non-null marker is not registered, or compare schema versions and explicitly reject a marker newer than this binary. Add a regression proving that a future/unknown marker is never replaced with an older one.

forkimenjeckayang (commented):
The requested `1.2.6 -> 1.3.0` migration is correct, idempotent, preserves configuration, and passes both focused and complete integration testing. The three comments  can be left as non-blocking clarification/hardening feedback. Of those, the future-marker behavior is the best candidate for a follow-up ticket before a second migration is introduced.
WDYT @IngridPuppet ?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/OID4VPMigrationManager.java
```diff
@@ -0,0 +1,154 @@
...
+     */
+    public void migrate(KeycloakSession session, RealmModel realm) {
+        if (config.shouldAutoCreateAuthFlowFor(realm.getName())) {
+            ensureOid4vpAuthFlowExists(realm);
+        }
+
+        if (realm.getFlowByAlias(OID4VPUserAuthEndpointBase.OID4VP_AUTH_FLOW) == null) {
+            logger.debugf(
+                    "Skipping migration for realm '%s': no oid4vp auth flow and the realm is not "
+                            + "in the managed-realms list",
```
Awambeng: Since we no longer migrate all realms, I think we should use `infof()` instead of `debugf()` so operators know that realms which aren't managed or don't have the OID4VP authenticator are being skipped.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/OID4VPMigrationManager.java
```diff
@@ -0,0 +1,154 @@
...
+        if (lastApplied == null) {
+            return 0;
+        }
+        for (int i = 0; i < migrations.size(); i++) {
+            if (lastApplied.equals(migrations.get(i).id())) {
+                return i + 1;
+            }
+        }
+        return 0;
+    }
```
Awambeng, line 159: Maybe we should add a sanity check to ensure migration IDs are unique. It would help prevent unexpected migration behavior.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/MigrationUtils.java
```diff
@@ -0,0 +1,112 @@
...
+     * <p>Configs whose ID is {@code null} (i.e. executions with no config attached) are skipped.
+     * Passing multiple IDs lets callers match both the current authenticator and any legacy
+     * aliases after a rename.
+     */
+    public static List<AuthenticatorConfigModel> configsInOid4vpFlow(RealmModel realm, String... providerIds) {
+        AuthenticationFlowModel flow = realm.getFlowByAlias(OID4VPUserAuthEndpointBase.OID4VP_AUTH_FLOW);
+        if (flow == null) {
+            return List.of();
+        }
+        Set<String> ids = Set.of(providerIds);
```
Awambeng: Should we avoid using `Set.of()` here since it throws an exception if duplicate elements are passed accidentally?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/MigrationUtils.java
Awambeng, line 1: Maybe worth adding unit tests for this utility since it might be reused by other tests. WDYT?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/steps/Migration_v1_3_0.java
```diff
@@ -0,0 +1,141 @@
...
+    public static final String ID = "v1.3.0";
+
+    public static final String LEGACY_AUTHENTICATOR = "sd-jwt-authenticator";
+    public static final String LEGACY_VCT_CONFIG_KEY = "vct";
+    public static final String LEGACY_KBJWT_MAX_AGE_CONFIG_KEY = "kbJwtMaxAge";
+
+    // Snapshot of authenticator identifiers as they existed in v1.3.0
+    public static final String CURRENT_AUTHENTICATOR = "oid4vp-authenticator";
+    public static final String CURRENT_CREDENTIAL_TYPES_CONFIG_KEY = "credentialTypes";
+    public static final String CURRENT_HOLDER_BINDING_PROOF_MAX_AGE_CONFIG_KEY = "holderBindingProofMaxAge";
```
Awambeng, line 49: Instead of redeclaring these strings, could we reuse the constants from `OID4VPAuthenticatorFactory`?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/steps/Migration_v1_3_0.java
```diff
@@ -0,0 +1,141 @@
...
+        if (renamed > 0) {
+            logger.infof("%s on %d authenticator config(s) in realm '%s'", operation, renamed, realm.getName());
+        } else {
+            logger.debugf(
+                    "No authenticator config in realm '%s' still carries legacy key '%s'", realm.getName(), legacyKey);
+        }
+    }
+
+    private void backfillFallbackToIsoSpecConfig(RealmModel realm) {
+        AtomicInteger updated = new AtomicInteger();
```
Awambeng: Any reason for using `AtomicInteger` here instead of a normal `int`? Since this is a regular loop, it looks like a simple counter would be sufficient.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/OID4VPConfig.java
```diff
@@ -2,21 +2,60 @@
...
+
+    /**
+     * Config key for the verbose-error-response toggle.
+     */
+    public static final String VERBOSE_ERRORS_CONFIG = "verbose-errors";
+
+    /**
+     * Config key for the certificate cache size.
+     * Or max entries for the authorization-request certificate cache.
+     */
```
Awambeng, line 32: Was this comment intentionally moved here?

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/OID4VPConfigTest.java
Awambeng, line 1: Should we add tests for duplicate and malformed managed realm entries to ensure they are handled correctly?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointFactory.java
```diff
@@ -47,70 +44,29 @@ public RealmResourceProvider create(KeycloakSession session) {
...
+        OID4VPMigrationManager migrationManager = new OID4VPMigrationManager(pluginConfig);
+
         factory.register(event -> {
             if (event instanceof PostMigrationEvent) {
                 logger.debugf("Migrating existing realms to add OpenID4VP user auth flow...");
-                KeycloakModelUtils.runJobInTransaction(
-                        factory, session -> session.realms().getRealmsStream().forEach(this::migrateRealmIfNecessary));
+                KeycloakModelUtils.runJobInTransaction(factory, session -> session.realms()
+                        .getRealmsStream()
+                        .forEach(realm -> migrationManager.migrate(session, realm)));
```
Awambeng, line 61: Should each realm migration run in its own transaction? Currently, a failure in one realm aborts the stream and prevents remaining realms from being migrated.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/OID4VPMigrationManagerTest.java
```diff
@@ -0,0 +1,198 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Consumer;
+import org.junit.jupiter.api.Test;
+import org.keycloak.Config;
+import org.keycloak.models.AuthenticationExecutionModel;
+import org.keycloak.models.AuthenticationFlowModel;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.RealmModel;
+
+class OID4VPMigrationManagerTest {
```
Awambeng, line 33: Should we add coverage for managed realms where the flow already exists? Also, it would be good to test the combination of existing flow handling with migration resume behavior, since these are currently only covered separately.

Awambeng (changes requested):

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/OID4VPMigrationManager.java
IngridPuppet: I agree it is best to reject unhealthy states for the current state of the plugin. We should take no risk. I added back the check (assertOid4vpAuthFlowIsHealthy). Thanks.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/OID4VPConfig.java
IngridPuppet: Yes, that's intentional. We'd like to prevent the flow from being created in realms that are not of target. The updates to the README document how to opt in.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/OID4VPMigrationManager.java
IngridPuppet: I agree it is best to reject unknown markers. I updated the code in this direction. This also aligns with the migration mechanism in Keycloak, which refuses to start when the marker is in the future. Thanks.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/OID4VPMigrationManager.java
IngridPuppet: Switched to infof. Thanks.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/OID4VPMigrationManager.java
IngridPuppet: I agree. Sanity check added.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/MigrationUtils.java
IngridPuppet: Thank you for the warning. I think it is okay to silently de-duplicate duplicated entries. This aligns with the intent of the method. The code was updated in this regard.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/MigrationUtils.java
IngridPuppet: I agree. Added.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/OID4VPConfigTest.java
IngridPuppet: I agree. Added.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/config/OID4VPConfig.java
IngridPuppet: Yes, for consistency with the other config options.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/OID4VPMigrationManagerTest.java
IngridPuppet: I agree. Added. `shouldNotAutoCreateFlowWhenManagedRealmAlreadyHasOne`

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/steps/Migration_v1_3_0.java
IngridPuppet: Updated to use a regular int. Thanks.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointFactory.java
IngridPuppet: Keycloak should anyway not start with a failed migration under any realm. I also think it is preferable to keep this all-or-none logic so realms are not left partially migrated, facilitating a downgrade.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/migration/steps/Migration_v1_3_0.java
IngridPuppet: I improved the comment explaining why we do this.

```java
    // These deliberately redeclare literals that OID4VPAuthenticatorFactory also exposes.
    // Referencing those constants would silently bind the migration to the *current* identifiers:
    // a later rename in the factory would make the migration look for the wrong strings in already-migrated
    // realms. Freezing the values here pins the migration to what was valid at ship time.
 ```

IngridPuppet:
> **Listener accumulation**: `postInit` registers a lambda on every factory initialization. Keycloak may call `postInit` on restarts or re-registration, causing duplicate migration runs per realm. Needs a guard flag.

Migrations are idempotent. The lastApplied marker is the flag.

Awambeng (approved):
Thanks @IngridPuppet, for resolving all my comments. LGTM! 🚀

forkimenjeckayang (approved):
LGTM

IngridPuppet:
Thank you for the reviews!

PR #142, opened 2026-07-29
feat: implement method to verify bearer token without consuming DPoP …

age-77:
…proof


IngridPuppet (approved):
I agree with the approach. Thanks.

PR #147, opened 2026-08-03
Add credential requirement groups for alternative OpenID4VP credentials

### LINKED ISSUE #119: Extend authentication profile configuration into groups of credential requirements
Follow up to:
- #88 

Support groups of credential requirements so the wallet only has to satisfy one option per group (e.g. accept a passport OR a national-ID as primary). 

Today every requirement is its own required CredentialSet in the DCQL query, so the wallet must present all of them, which is problematic if there is need to support one or the other format for a given type of credential requirement.

ETA: 1 day

Ogenbertrand:
This adds `credentialGroups` to authentication profiles so a verifier can request alternatives like passport OR national ID while keeping the existing flat credential configuration valid. The DCQL query, authorization response validation, and authenticator flow now resolve the selected option per group and authenticate with the credential actually presented by the wallet.

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
@@ -66,12 +66,19 @@ public class OID4VPAuthenticator implements Authenticator {
         logger.info("Authenticating with OID4VPAuthenticator");
 
         AuthenticationProfile profile = getAuthenticationProfile(context);
-        CredentialRequirement primaryCredential = profile.getPrimaryCredential();
 
         AuthorizationContext authContext = getAuthorizationContext(authSession);
         AuthRequirements authRequirements = new AuthRequirements(context.getAuthenticatorConfig());
 
         Map<String, String> presentedTokens = getPresentedTokens(authSession);
+        CredentialRequirement primaryCredential;
+        try {
+            primaryCredential = profile.getPresentedPrimaryCredential(presentedTokens.keySet());
+        } catch (IllegalStateException e) {
+            failRejectingPresentedCredential(context, e.getMessage(), e);
+            return;
+        }
+
         String primaryToken = presentedTokens.get(primaryCredential.getId());
         if (StringUtil.isBlank(primaryToken)) {
             failRejectingPresentedCredential(
@@ -186,8 +193,7 @@ public class OID4VPAuthenticator implements Authenticator {
 
             String token = supportingTokens.get(credential.getId());
             if (StringUtil.isBlank(token)) {
-                throw new VerificationException(String.format(
-                        "Supporting credential '%s' is missing from the presentation", credential.getId()));
+                continue;
             }
 
             CredentialVerifier supportingVerifier = resolveVerifier(authSession, credential.getId());
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
@@ -14,6 +14,7 @@ import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.profile.CredentialRe
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.profile.OID4VPProfileConfig;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.service.AuthenticationSessionStore;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.utils.ErrorResponseSanitizer;
+import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.utils.TransactionDataSupport;
 import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
 import java.io.IOException;
@@ -133,8 +134,10 @@ public class OID4VPAuthenticator implements Authenticator {
         JsonNode primaryClaims;
         try {
             primaryClaims = primaryVerifier.verifyCredential(ctx, primaryCredentialReq, primaryToken);
+            TransactionDataSupport.requireCredentialIdInAllEntries(
+                    ctx.authorizationContext().getRequestObject().getTransactionData(), primaryCredentialReq.getId());
             primaryVerifier.validateTransactionData(ctx, primaryToken);
-        } catch (VerificationException | IllegalStateException e) {
+        } catch (VerificationException | IllegalArgumentException | IllegalStateException e) {
             String msg = "Primary credential verification failed";
             failRejectingPresentedCredential(ctx, String.format("%s: %s", msg, e.getMessage()));
             throw new VerificationException(msg, e);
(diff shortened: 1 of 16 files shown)


forkimenjeckayang:
Please address conflicts so that I can check @Ogenbertrand

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
```diff
@@ -185,7 +189,7 @@ void applyBindingRules(
             Context ctx, AuthenticatingUser authUser, CredentialRequirement credentialReq, JsonNode claims)
             throws VerificationException {
         KeycloakSession session = ctx.authenticationFlowContext().getSession();
-        CredentialRequirement primaryCredentialReq = ctx.authenticationProfile().getPrimaryCredential();
+        CredentialRequirement primaryCredentialReq = getPresentedPrimaryCredential(ctx);
```
forkimenjeckayang: > The full test suite currently fails in four existing `SdJwtPrimaryBindingTest` cases. `applyBindingRules()` now resolves a presented primary before examining the rules, including when there are no binding rules or when the only rule is `claim_equals_user_attribute`, neither of which needs a primary credential token. Those unit contexts correctly did not need to populate presented-token state for these operations, and now fail with `Presentation must contain exactly one primary credential: default`.
>

 Please resolve the presented primary lazily only inside the `CLAIM_EQUALS_PRIMARY_CLAIM` branch, or pass the already selected primary explicitly from the caller. Then rerun the complete suite. The current failures are `passesWhenClaimMatchesUserAttribute`, `passesWhenNoBindingRules`, `throwsWhenClaimDoesNotMatchUserAttribute`, and `throwsWhenPrimaryUsesPrimaryClaimRule`.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
forkimenjeckayang, line 139: > This correctly selects the primary credential that the Wallet presented, but transaction-data preparation and validation still assume the first DCQL credential. `AuthorizationRequestService` lines 270-280 calls `prepareWireEntries(..., dcqlQuery.getCredentials().getFirst().getId())`, so generated `transaction_data.credential_ids` references only that first credential. `validateTransactionData()` then checks the hashes but never verifies that `primaryCredentialReq.getId()` is listed in the transaction-data object.
>

> With alternatives such as `passport OR national-id`, the Wallet can select `national-id`, sign matching transaction-data hashes with it, and this path accepts the presentation even though the signed request says only `passport` may authorize the transaction. Final OID4VP 1.0 requires `credential_ids` to identify the Credential Queries that may authorize each transaction, and where several are listed the Wallet uses one referenced credential.
>

Please prepare each transaction-data entry with all permitted primary alternatives (or reject grouped primary profiles when transaction data is configured until that is supported), and verify that the selected primary ID is listed before accepting its transaction-data proof. The same first-credential assumption in `VerifierInfoSupport.build(...)` should be reviewed for grouped profiles. Add positive and negative tests selecting a non-first primary with transaction data.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -154,12 +160,32 @@ private HashMap<String, String> extractPresentedTokens(
             throw failInvalidVpToken("Presented vp_token map is missing", authContext, store);
         }
 
-        for (CredentialRequirement credential : profile.getCredentials()) {
-            var credentialTokens = vpTokenMap.get(credential.getId());
+        Map<String, CredentialRequirement> credentialsById = profile.getCredentials().stream()
+                .collect(Collectors.toMap(CredentialRequirement::getId, credential -> credential));
+        Set<String> selectedCredentialIds =
+                selectPresentedCredentialIds(vpTokenMap.keySet(), profile, authContext, store);
```
forkimenjeckayang: > The accepted credential combination is being derived from the current `AuthenticationProfile`, which `OID4VPUserAuthEndpoint` reparses from realm configuration when the response arrives. The immutable request stored in `AuthorizationContext` already contains the exact signed `dcql_query` that the Wallet answered. If an administrator changes the profile while a request is in flight, this code can accept a combination allowed by the new profile but not by the signed request, or reject a response that correctly satisfies the original request.
>

Please make `authContext.getRequestObject().getDcqlQuery().getCredentialSets()` the single source of truth for response-combination validation. The profile can still map an authorized credential ID to local verifier, role, trust, and user-recovery configuration, but it must not replace the persisted protocol policy. This also removes duplicated group interpretation between request generation and response processing. Add a regression where the profile changes after request creation and the response is still judged only against the stored DCQL query.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -173,13 +199,61 @@ private HashMap<String, String> extractPresentedTokens(
...
+            AuthorizationContext authContext,
+            AuthenticationSessionStore store) {
+        Set<String> selectedCredentialIds = new HashSet<>();
+
+        for (CredentialRequirementGroup group : profile.getEffectiveCredentialGroups()) {
+            List<List<String>> selectedOptions = group.getOptions().stream()
+                    .filter(option -> presentedCredentialIds.containsAll(option))
+                    .toList();
+
+            if (group.isRequired() && selectedOptions.size() != 1) {
```
forkimenjeckayang: > Final 1.0 requires the returned credential combination to match one acceptable option; it does not require that only one option be a subset of the global response. For a valid set with `options: [["pid"], ["pid", "residence"]]`, a Wallet returning `pid` and `residence` satisfies the second option, but `containsAll()` matches both and this code rejects it with `Found: 2`. The same problem affects duplicate or subset-related optional alternatives, all of which the current profile validator permits and DCQL does not prohibit.
>

Please validate the complete presented combination across required and optional sets, selecting options whose union accounts for the returned credential IDs, instead of requiring one subset match independently per group. A small shared DCQL response-selection component operating on the persisted `credential_sets` would keep this logic out of the profile model. Add tests for multi-ID options, subset/overlapping options, optional sets, and credentials shared by more than one set.

forkimenjeckayang (changes requested):
a few comments @Ogenbertrand  , please check

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
Ogenbertrand: Thanks fixed

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
Ogenbertrand: done

forkimenjeckayang (approved):
No further concderns but i'll like @IngridPuppet to check also

IngridPuppet (approved):
Just skimmed through. Should be okay.

PR #134, opened 2026-07-22
Refactor DCQL capabilities for format-specific validation

forkimenjeckayang:
This PR improves the DCQL capability stack so each credential format can define its own validation rules.

## What changed

- Separated common DCQL validation from format-specific validation.
- Added a clear capability contract for SD-JWT, mDoc, and future formats.
- Added deterministic capability lookup by credential format.
- Moved SD-JWT-specific rules into the SD-JWT capability.
- Added mDoc query validation for:
  - required `doctype_value`;
  - claim paths containing exactly a namespace and element identifier.
- Removed the hard-coded mDoc check from response processing.
- Added validation for generated DCQL authorization requests.
- Added tests
## Why

Previously, the shared validator contained SD-JWT-specific assumptions, and response processing needed to know which formats supported pre-validation.

Now, each format owns its validation rules and reports whether it supports presentation pre-validation. Implementing #117 should therefore be mostly limited to `MdocDcqlCredentialCapability`.

Closes #120

PR #149, opened 2026-08-05
Implement DCQL presentation pre-validation for mDoc credential queries

### LINKED ISSUE #117: Implement DCQL presentation pre-validation for mDoc credential queries
Follow up to:
- #88 

As it is done for presented SD-JWTs, quickly validate that mDoc presentations satisfy corresponding DCQL credential queries.

ETA: 1 day

forkimenjeckayang:
Add format-specific DCQL presentation pre-validation for `mso_mdoc` credential queries using the capability architecture introduced in #134.

The verifier now rejects mDoc presentations early when the presented `DeviceResponse` does not satisfy the corresponding DCQL credential query.

## Implementation

- Enable presentation pre-validation for the mDoc DCQL capability.
- Decode the base64url-encoded mDoc `DeviceResponse`.
- Require exactly one presented Document for each DCQL credential response.
- Validate the presented Document `docType` against `meta.doctype_value`.
- Extract issuer-signed data-element paths from `IssuerSigned.nameSpaces`.
- Validate exact `[namespace, elementIdentifier]` claim paths.
- Require every requested claim when `claim_sets` is absent.
- Accept any complete alternative when `claim_sets` is present.
- Treat claim `values` as a best-effort privacy hint rather than a security constraint, as required by OpenID4VP Final 1.0.
- Preserve cryptographic verification, digest validation, trust validation, validity checks, and device binding in the existing mDoc credential verifier.

closes #117

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapability.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapability.java
@@ -1,15 +1,32 @@
 package io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.dcql;
 
+import com.authlete.cbor.CBORItem;
+import com.authlete.cbor.CBORItemList;
+import com.authlete.cbor.CBORPair;
+import com.authlete.cbor.CBORPairList;
+import com.authlete.cbor.CBORString;
 import com.authlete.cose.constants.COSEAlgorithms;
+import io.github.adorsysgis.keycloak.protocol.oid4vc.mdoc.CborUtil;
+import io.github.adorsysgis.keycloak.protocol.oid4vc.mdoc.MdocConstants;
+import io.github.adorsysgis.keycloak.protocol.oid4vc.mdoc.MdocEncodingException;
+import io.github.adorsysgis.keycloak.protocol.oid4vc.mdoc.MdocParser;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.authenticator.CredentialFormat;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model.ClientMetadata;
+import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model.dcql.Claim;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model.dcql.Credential;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model.prex.MdocGenericFormat;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
+import java.util.Set;
 import org.keycloak.common.VerificationException;
 
 /**
  * DCQL capability for {@code mso_mdoc} (ISO/IEC 18013-5) credentials.
+ *
+ * <p>Pre-validates the presented DeviceResponse document type and the presence of requested
+ * namespace/data-element paths. Cryptographic, digest, validity, trust, and device-binding checks remain in the
+ * mDoc credential verifier.
  *
  * <p>Contributes VP format metadata with COSE algorithm identifiers for
  * issuer authentication and device authentication per OpenID4VP Final 1.0
@@ -24,13 +41,24 @@ public final class MdocDcqlCredentialCapability implements DcqlCredentialCapabil
 
     @Override
     public boolean supportsPresentationPreValidation() {
-        return false;
+        return true;
     }
 
     @Override
     public void validatePresentation(Credential credential, String presentedToken) throws VerificationException {
-        throw new UnsupportedOperationException(
-                "DCQL presentation validation is not yet supported for mso_mdoc credentials");
+        if (!format().equals(credential.getFormat())) {
+            throw new VerificationException(
+                    "Unsupported dcql_query credential format for presentation validation: " + credential.getFormat());
+        }
+
+        PresentedMdoc presentation = parsePresentation(presentedToken);
+        String expectedDocType = credential.getMeta().getDoctypeValue();
+        if (!Objects.equals(expectedDocType, presentation.docType())) {
+            throw new VerificationException(
+                    "Presented mDoc docType does not match meta.doctype_value: " + presentation.docType());
+        }
+
+        validateRequestedClaims(credential, presentation.claimPaths());
     }
 
     @Override
@@ -46,4 +74,96 @@ public final class MdocDcqlCredentialCapability implements DcqlCredentialCapabil
         format.setDeviceAuthAlgValues(coseAlgorithms);
         vpFormat.setMsoMdoc(format);
     }
+
+    private static PresentedMdoc parsePresentation(String presentedToken) throws VerificationException {
+        CBORPairList deviceResponse;
+        try {
+            deviceResponse = MdocParser.parseBase64Url(presentedToken);
+        } catch (MdocEncodingException e) {
+            throw new IllegalArgumentException("Failed to parse presented mDoc device response", e);
+        }
+
+        CBORPair documentsEntry = deviceResponse.findByKey(MdocConstants.L_DOCUMENTS);
+        if (documentsEntry == null) {
+            throw new VerificationException("Presented mDoc response is missing the documents field");
+        }
+
+        CBORItemList documents = (CBORItemList) documentsEntry.getValue();
+        if (documents.getItems().size() != 1) {
+            throw new VerificationException("Presented mDoc response must contain exactly one document, but contained "
+                    + documents.getItems().size());
+        }
+
+        CBORPairList document = (CBORPairList) documents.getItems().getFirst();
+        String docType =
+                ((CBORString) document.findByKey(MdocConstants.L_DOC_TYPE).getValue()).getValue();
+        return new PresentedMdoc(docType, extractClaimPaths(document));
+    }
+
+    private static Set<ClaimPath> extractClaimPaths(CBORPairList document) {
+        CBORPairList issuerSigned =
+                (CBORPairList) document.findByKey(MdocConstants.L_ISSUER_SIGNED).getValue();
+        CBORPair namespacesEntry = issuerSigned.findByKey(MdocConstants.L_NAME_SPACES);
+        if (namespacesEntry == null) {
+            return Set.of();
+        }
+
+        Set<ClaimPath> claimPaths = new HashSet<>();
+        CBORPairList namespaces = (CBORPairList) namespacesEntry.getValue();
+        for (CBORPair namespace : namespaces.getPairs()) {
+            String namespaceIdentifier = CborUtil.asString(namespace.getKey());
+            CBORItemList issuerSignedItems = (CBORItemList) namespace.getValue();
+            for (CBORItem issuerSignedItem : issuerSignedItems.getItems()) {
+                CBORPairList item = (CBORPairList) CborUtil.unwrap(issuerSignedItem);
+                String elementIdentifier = ((CBORString) item.findByKey(MdocConstants.L_ELEMENT_IDENTIFIER)
+                                .getValue())
+                        .getValue();
+                claimPaths.add(new ClaimPath(namespaceIdentifier, elementIdentifier));
+            }
+        }
+        return Set.copyOf(claimPaths);
+    }
+
+    /**
+     * Validates claim-path presence and claim-set alternatives. Claim {@code values} are intentionally not enforced:
+     * OpenID4VP Final 1.0 defines them as a best-effort wallet privacy hint that verifiers must not treat as a
+     * security control.
+     */
+    private static void validateRequestedClaims(Credential credential, Set<ClaimPath> presentedClaimPaths)
+            throws VerificationException {
+        List<Claim> claims = credential.getClaims();
+        if (claims == null || claims.isEmpty()) {
+            return;
+        }
+
+        List<List<String>> claimSets = credential.getClaimSets();
+        if (claimSets == null || claimSets.isEmpty()) {
+            for (Claim claim : claims) {
+                if (!isPresented(claim, presentedClaimPaths)) {
+                    throw new VerificationException(
+                            "Presented mDoc does not satisfy DCQL claim path: " + claim.getPath());
+                }
+            }
+            return;
+        }
+
+        boolean satisfiesAnyClaimSet = claimSets.stream()
+                .anyMatch(option -> option.stream().allMatch(claimId -> claims.stream()
+                        .filter(claim -> Objects.equals(claimId, claim.getId()))
+                        .anyMatch(claim -> isPresented(claim, presentedClaimPaths))));
+        if (!satisfiesAnyClaimSet) {
+            throw new VerificationException("Presented mDoc does not satisfy any DCQL claim_sets option");
+        }
+    }
+
+    private static boolean isPresented(Claim claim, Set<ClaimPath> presentedClaimPaths) {
+        List<String> path = claim.getPath();
+        return path != null
+                && path.size() == 2
+                && presentedClaimPaths.contains(new ClaimPath(path.getFirst(), path.get(1)));
+    }
+
+    private record PresentedMdoc(String docType, Set<ClaimPath> claimPaths) {}
+
+    private record ClaimPath(String namespace, String elementIdentifier) {}
 }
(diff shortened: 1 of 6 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapability.java
```diff
@@ -24,13 +41,24 @@ public String format() {
...
+        if (!format().equals(credential.getFormat())) {
+            throw new VerificationException(
+                    "Unsupported dcql_query credential format for presentation validation: " + credential.getFormat());
+        }
+
+        PresentedMdoc presentation = parsePresentation(presentedToken);
+        String expectedDocType = credential.getMeta().getDoctypeValue();
+        if (!Objects.equals(expectedDocType, presentation.docType())) {
+            throw new VerificationException(
+                    "Presented mDoc docType does not match meta.doctype_value: " + presentation.docType());
```
Awambeng: Could you please add the expected type to the error message? This will make debugging easier.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapability.java
```diff
@@ -46,4 +74,96 @@ public void contributeVpFormatsSupported(ClientMetadata.VpFormat vpFormat, List<
...
+                            "Presented mDoc does not satisfy DCQL claim path: " + claim.getPath());
+                }
+            }
+            return;
+        }
+
+        boolean satisfiesAnyClaimSet = claimSets.stream()
+                .anyMatch(option -> option.stream().allMatch(claimId -> claims.stream()
+                        .filter(claim -> Objects.equals(claimId, claim.getId()))
+                        .anyMatch(claim -> isPresented(claim, presentedClaimPaths))));
```
Awambeng: This is a bit hard to follow with the nested streams. In the SD-JWT capability we precompute a `Map<claimId, satisfied>` and then just do lookups, which I found much easier to read. I think the same approach would work well here too.

Also, we're now duplicating the `claims`/`claim_sets` evaluation logic across both capabilities. Since they share the same model, I'd consider extracting this into a small shared helper instead of maintaining two copies.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapability.java
```diff
@@ -46,4 +74,96 @@ public void contributeVpFormatsSupported(ClientMetadata.VpFormat vpFormat, List<
...
+        if (!satisfiesAnyClaimSet) {
+            throw new VerificationException("Presented mDoc does not satisfy any DCQL claim_sets option");
+        }
+    }
+
+    private static boolean isPresented(Claim claim, Set<ClaimPath> presentedClaimPaths) {
+        List<String> path = claim.getPath();
+        return path != null
+                && path.size() == 2
+                && presentedClaimPaths.contains(new ClaimPath(path.getFirst(), path.get(1)));
```
Awambeng: This assumes mDoc claim paths are always [namespace, element], but that's not obvious from the code. If a path has a different shape, we just return false, which later results in the generic "does not satisfy DCQL claim path" error even though the path wasn't actually evaluated.

Could we either validate the path shape explicitly (with a clearer error like "mDoc claims must be [namespace, element]") or document this assumption at the start of the method? Right now it's a bit implicit.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapability.java
```diff
@@ -46,4 +74,96 @@ public void contributeVpFormatsSupported(ClientMetadata.VpFormat vpFormat, List<
...
+            CBORItemList issuerSignedItems = (CBORItemList) namespace.getValue();
+            for (CBORItem issuerSignedItem : issuerSignedItems.getItems()) {
+                CBORPairList item = (CBORPairList) CborUtil.unwrap(issuerSignedItem);
+                String elementIdentifier = ((CBORString) item.findByKey(MdocConstants.L_ELEMENT_IDENTIFIER)
+                                .getValue())
+                        .getValue();
+                claimPaths.add(new ClaimPath(namespaceIdentifier, elementIdentifier));
+            }
+        }
+        return Set.copyOf(claimPaths);
```
Awambeng: The `HashSet` never escapes this method, so the copy buys nothing. Either return  `claimPaths`  or just collect directly into an immutable set.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapabilityTest.java
```diff
@@ -16,6 +26,125 @@ void formatReturnsMsoMdoc() {
...
+    void rejectsUnsupportedCredentialFormat() throws Exception {
+        String token = mdocToken(Map.of(NAMESPACE, Map.of("given_name", "Alice")), DOC_TYPE);
+        Credential credential = credentialWithClaims(claim("given-name", NAMESPACE, "given_name"));
+        credential.setFormat(VCFormat.SD_JWT_VC);
+
+        VerificationException error =
+                assertThrows(VerificationException.class, () -> capability.validatePresentation(credential, token));
+        assertEquals(
+                "Unsupported dcql_query credential format for presentation validation: " + VCFormat.SD_JWT_VC,
+                error.getMessage());
```
Awambeng: Should we hard-code the expected message here instead of building it from `VCFormat.SD_JWT_VC`. Otherwise the test is using the same value as the production code, so it won't catch regressions in the message.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapabilityTest.java
Awambeng, line 1: Looks like we're still missing coverage for a few branches:

- missing documents
- 0 or multiple documents
- missing nameSpaces
- invalid claim path length

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapabilityTest.java
```diff
@@ -16,6 +26,125 @@ void formatReturnsMsoMdoc() {
...
+        String token = mdocToken(Map.of(NAMESPACE, Map.of("given_name", "Alice")), DOC_TYPE);
+        Claim givenName = claim("given-name", NAMESPACE, "given_name");
+        givenName.setValues(List.of("Bob"));
+        Credential credential = credentialWithClaims(givenName);
+
+        assertDoesNotThrow(() -> capability.validatePresentation(credential, token));
+    }
+
+    @Test
+    void rejectsMalformedPresentation() {
```
Awambeng, line 155: Could we also add a case where the CBOR is valid but the document is invalid (e.g. missing docType)? That seems like a more realistic failure mode than "not-cbor" and exercises a different code path.

Awambeng (changes requested):

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapability.java
forkimenjeckayang: Agreed. I extracted the format-neutral `claims/claim_sets` evaluation into `DcqlClaimSelectionValidator`. It precomputes a `Map<claimId, satisfied>` for claim-set evaluation, while each capability still supplies its format-specific claim evaluator. This removes the nested streams and duplicated selection logic without changing SD-JWT or mDoc validation behavior.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapability.java
forkimenjeckayang: added explicit validation with a clearer error message.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapabilityTest.java
```diff
@@ -145,6 +151,29 @@ void rejectsMalformedPresentation() {
...
+    void rejectsValidCborButMissingDocType() throws Exception {
+        CBORPairList issuerSigned =
+                new CBORPairList(new CBORPair(new CBORString(MdocConstants.L_NAME_SPACES), new CBORPairList()));
+
+        CBORPairList document =
+                new CBORPairList(new CBORPair(new CBORString(MdocConstants.L_ISSUER_SIGNED), issuerSigned));
+
+        CBORItemList documents = new CBORItemList(List.of(document));
+        CBORPairList deviceResponse =
+                new CBORPairList(new CBORPair(new CBORString(MdocConstants.L_DOCUMENTS), documents));
```
Awambeng: These tests pass for the wrong reason. The built CBOR only contains  documents, but the schema requires  `version` / `status`  at the root. So `MdocParser` fails at root schema validation and the `IllegalArgumentException("Failed to parse...")`  you assert never has anything to do with the missing  docType. 

The actual branch you're naming ( `docType`  lookup in  `parsePresentation` ) is never reached. Add  `version` / `status`  to the wrapper so the input is otherwise valid, then this test actually tests what it says.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapabilityTest.java
```diff
@@ -145,6 +151,29 @@ void rejectsMalformedPresentation() {
...
+
+        String validCborButInvalidMdoc =
+                java.util.Base64.getUrlEncoder().withoutPadding().encodeToString(deviceResponse.encode());
```
Awambeng: please avoid using FQN and import this at the top of the file.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapabilityTest.java
```diff
@@ -212,4 +316,40 @@ private static Claim claim(String id, String namespace, String elementIdentifier
...
+        CBORPairList document = new CBORPairList(
+                new CBORPair(new CBORString(MdocConstants.L_DOC_TYPE), new CBORString(DOC_TYPE)),
+                new CBORPair(new CBORString(MdocConstants.L_ISSUER_SIGNED), issuerSigned));
+
+        CBORItemList documents = new CBORItemList(List.of(document));
+
+        CBORPairList deviceResponse =
+                new CBORPairList(new CBORPair(new CBORString(MdocConstants.L_DOCUMENTS), documents));
+
+        return java.util.Base64.getUrlEncoder().withoutPadding().encodeToString(deviceResponse.encode());
```
Awambeng: same here.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapabilityTest.java
```diff
@@ -212,4 +316,40 @@ private static Claim claim(String id, String namespace, String elementIdentifier
         claim.setPath(List.of(namespace, elementIdentifier));
         return claim;
     }
+
+    private static String buildDeviceResponseWithDocumentCount(int documentCount) throws Exception {
```
Awambeng: The test inputs are missing `version` and `status`, so schema validation fails before reaching the intended branch. After adding them, the expected exception should be `VerificationException` rather than `IllegalArgumentException`

See https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/149/changes/3f717da14292c283c89592ccfe18b0aafb9d03e5#r3736159446

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapabilityTest.java
```diff
@@ -212,4 +316,40 @@ private static Claim claim(String id, String namespace, String elementIdentifier
...
+        List<CBORPairList> documentList = new java.util.ArrayList<>();
```
Awambeng: Avoid using FQN and import the class directly.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapabilityTest.java
```diff
@@ -185,6 +214,81 @@ void contributeVpFormatsSupportedHandlesEdDSA() {
...
+    @Test
+    void rejectsPresentationWithMultipleDocuments() throws Exception {
+        IllegalArgumentException error = assertThrows(
+                IllegalArgumentException.class,
+                () -> capability.validatePresentation(credentialWithClaims(), buildDeviceResponseWithDocumentCount(2)));
+        assertTrue(error.getMessage().contains("Failed to parse"));
+    }
+
+    @Test
+    void handlesPresentationWithMissingNamespaces() throws Exception {
```
Awambeng, line 256: Shouldn't this be `assertDoesNotThrow`? nameSpaces isn't required by the schema, so a valid response without it should be accepted. Also, could you check whether `version` and `status` are needed here? As it stands, the test fails schema validation before reaching the nameSpaces handling logic.

Awambeng (changes requested):

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapabilityTest.java
```diff
@@ -55,4 +222,157 @@ void contributeVpFormatsSupportedHandlesEdDSA() {
...
+    @Test
+    void rejectsPresentationWithMultipleDocuments() throws Exception {
+        VerificationException error = assertThrows(
+                VerificationException.class,
+                () -> capability.validatePresentation(credentialWithClaims(), buildDeviceResponseWithDocumentCount(2)));
+        assertTrue(error.getMessage().contains("exactly one document"));
+    }
+
+    @Test
+    void handlesPresentationWithMissingNamespaces() throws Exception {
```
Awambeng, line 260: This test could be strengthened by also requesting a claim and asserting the failure. WDYT?

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapabilityTest.java
```diff
@@ -16,6 +38,151 @@ void formatReturnsMsoMdoc() {
...
+
+    @Test
+    void rejectsMalformedPresentation() {
+        Credential credential = credentialWithClaims(claim("given-name", NAMESPACE, "given_name"));
+
+        assertThrows(IllegalArgumentException.class, () -> capability.validatePresentation(credential, "not-cbor"));
+    }
+
+    @Test
+    void rejectsValidCborButMissingDocType() throws Exception {
```
Awambeng, line 162: `rejectsValidCborButMissingDocType` name overpromises slightly — the document is missing both  docType  and  deviceSigned , so it fails schema validation for two reasons. The intent is genuinely tested; isolating  docType  would require  deviceSigned  present.

Awambeng (commented):
Thanks for addressing the comments, @forkimenjeckayang. However, could you please also check the NPE comment from Opencode and resolve it?

forkimenjeckayang:
> Thanks for addressing the comments, @forkimenjeckayang. However, could you please also check the NPE comment from Opencode and resolve it?

`MdocParser.parseBase64Url()` performs full schema validation before returning. The schema requires each Document to contain a string `docType`, so a missing `docType` raises `MdocEncodingException` before this access is reached. The capability then translates that parsing failure to `IllegalArgumentException`, per `DcqlCredentialCapability`’s contract. The isolated regression test also asserts that the underlying schema error specifically references docType. Therefore, the reported NPE is not reachable and an additional null check here would duplicate parser validation.

Awambeng (approved):

PR #152, opened 2026-08-10
Hide session identity profiles from login page

### LINKED ISSUE #151: Hide login buttons associated with session-identity authentication profiles
Authentication profiles may be defined to source the identity of the authenticating user from session rather than from credential. Such authentication profiles are only meant to used as part of other flows. The login page must not display login buttons for them.

ETA: 4 hours

Ogenbertrand:
Closes: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/151

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/AuthenticationProfile.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/AuthenticationProfile.java
@@ -133,6 +133,16 @@ public class AuthenticationProfile {
         return primaryCredentials.getFirst();
     }
 
+    /**
+     * Session-identity profiles are internal flow steps and must not be offered as standalone login choices.
+     */
+    public boolean isSessionIdentityProfile() {
+        return credentials != null
+                && credentials.stream()
+                        .filter(CredentialRequirement::isPrimary)
+                        .anyMatch(CredentialRequirement::isSessionIdentity);
+    }
+
     public CredentialRequirement getCredential(String credentialId) {
         return credentials.stream()
                 .filter(credential -> credential.getId().equals(credentialId))
--- a/src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPLoginActionsServiceTest.java
+++ b/src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPLoginActionsServiceTest.java
@@ -171,43 +171,12 @@ public class OID4VPLoginActionsServiceTest extends OID4VPBaseUserAuthEndpointTes
         var originalConfig = getAuthenticatorConfig();
         try {
             var updatedConfig = getAuthenticatorConfig();
-            updatedConfig.getConfig().put(PROFILES_CONFIG, """
-                    [
-                      {
-                        "id": "default",
-                        "displayCta": { "en": "Sign in with a wallet" },
-                        "credentials": [
-                          {
-                            "id": "identity",
-                            "role": "primary",
-                            "credentialTypes": ["https://credentials.example.com/identity_credential"],
-                            "claims": ["sub", "username"]
-                          }
-                        ]
-                      },
-                      {
-                        "id": "issuance-step",
-                        "displayCta": { "en": "Presentation during issuance" },
-                        "enabledForClients": ["test-app"],
-                        "credentials": [
-                          {
-                            "id": "pid",
-                            "role": "primary",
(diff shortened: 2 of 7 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/AuthenticationProfile.java
```diff
@@ -133,6 +133,16 @@ public CredentialRequirement getPresentedPrimaryCredential(Collection<String> pr
...
     }
 
+    /**
+     * Session-identity profiles are internal flow steps and must not be offered as standalone login choices.
+     */
+    public boolean isSessionIdentityProfile() {
+        return credentials != null
+                && credentials.stream()
+                        .filter(CredentialRequirement::isPrimary)
+                        .anyMatch(CredentialRequirement::isSessionIdentity);
```
forkimenjeckayang, line 143: > This classifies the whole profile as session-only when **any** possible primary credential uses `identitySource=session`. Credential groups currently allow alternative primary credentials, and profile validation validates each primary independently without requiring all alternatives to use the same identity source. The authenticator also chooses the identity source from the primary credential actually presented.
>
> A valid group such as `[["wallet-login"], ["issuance-proof"]]`, where the first primary uses credential identity and the second uses session identity, is therefore hidden entirely even though it contains a standalone login alternative.
>
Please make this invariant explicit. The safest model is to reject mixed primary identity sources during profile validation, making a profile unambiguously either standalone-login or session-only. If mixed profiles are intentionally supported, the login path needs to expose only credential-identity alternatives rather than hiding the complete profile or offering a session-only alternative. Please add a regression test for alternative primary credentials with different identity sources.

forkimenjeckayang (commented):
Please check when you get the chance @Ogenbertrand

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPLoginActionsServiceTest.java
```diff
@@ -164,6 +166,69 @@ public void shouldProvisionOid4vpLogin_WhenOidcAuthorizeWithoutPkce() throws Exc
...
+                                "type": "claim_equals_user_attribute",
+                                "credentialClaim": "family_name",
+                                "userAttribute": "lastName"
+                              }
+                            ]
+                          }
+                        ]
+                      }
+                    ]
+                    """);
```
IngridPuppet: Could you relocate this to the existing AuthenticationProfileSamples class?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/AuthenticationProfile.java
IngridPuppet: I think it is best to avoid such profiles mixing credential and session-identity sources. I can't find a valid reason why we would want to allow that. I think all possible primary credentials MUST have the same identity source configuration. And this code should be fine, rejecting all profiles with any session-identity credential requirement; it is simpler to reason about it that way. What's your point of view, @Ogenbertrand ?

IngridPuppet (commented):
Thank you for the PR. My comments are non-blocking.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/AuthenticationProfile.java
Ogenbertrand: Thanks to you all for the contribution. Mixing credential and session identity sources across alternative primary credentials makes the profile somehow unclear, the login UI would not know whether the profile is a standalone login option or an internal session-bound flow. I think we should make the invariant explicit in profile validation, all possible primary credentials in one authentication profile must use the same identitySource. Then the current login filtering remains simple and correct profiles with any/session primary are treated as session-only and hidden from the login page. I’ll add a regression test for mixed primary alternatives to ensure such profiles are rejected.

WDYT @IngridPuppet @forkimenjeckayang ?

IngridPuppet: I agree

Ogenbertrand: Done.

forkimenjeckayang (approved):
LGTM

IngridPuppet (approved):

PR #157, opened 2026-08-17
 fix(oid4vp): make primary credential identity claims configurable

### LINKED ISSUE #153: mDoc: primary credential hard-requires `sub`/`username` claims, rejecting valid ISO 18013-5 mDLs
> reproduced on **both** verifier plans, HAIP
> (plan `fimWPD2hbMjzY`, test `85HfnFfB0oPh1ca`) and plain_vp (plan `EK6A82IpBZQt0`, test
> `KJdGzVitCBLau17`),  same suite failure and same Keycloak log line on each.

***Type:*** bug (compliance / interoperability)
***Suggested labels:*** `bug`, `oid4vp`, `mDoc`

#### Summary

During OpenID4VP conformance testing (mDoc/ISO mDL verifier profile), the verifier
**rejects a valid, correctly signed, trusted ISO 18013-5 mDL** because the plugin requires
the primary credential to contain `sub` and `username` claims, which are **not defined in
the ISO 18013-5 mDL** and therefore cannot be present in a standard mDL.

#### Evidence

Conformance suite: plan `fimWPD2hbMjzY` (suite 5.2.2), module
`oid4vp-1final-verifier-happy-flow`, test `85HfnFfB0oPh1ca`,
variant `credential_format=iso_mdl, client_id_prefix=x509_hash,
request_method=request_uri_signed, vp_profile=haip, response_mode=direct_post.jwt`.

Suite failure:

> **OID4VP-1FINAL-8.2 EnsureHttpStatusCodeIs200:** `https://.../realms/oid4vp-demo/oid4vp-auth/response/<requestId>` endpoint returned an unexpected http status

Image

Image

**Confirmation on second run:** reproduced identically on a **plain (non-HAIP)**
verifier plan — plan `EK6A82IpBZQt0`, test `KJdGzVitCBLau17`,
`vp_profile=plain_vp` → same `OID4VP-1FINAL-8.2` failure and the same Keycloak log line
(`Presented mDoc does not satisfy DCQL claim path: [org.iso.18013.5.1, sub]`, request id
`NQtHUulnwrPQRQKZrSRk-VVQ.uspR3GEw7O0` matching the pasted link). This confirms the issue is
**independent of the certification profile**.

#### Root cause

1. `OID4VPProfileConfig.validatePrimaryCredential()` **hard-requires** every
   credential-identity primary credential to request `sub` and `username`:
   `"OpenID4VP primary credential must request sub and username"`.
2. The DCQL pre-validation (`MdocDcqlCredentialCapability.validatePresentation` →
   `DcqlClaimSelectionValidator`) then rejects any presented mDoc that does not contain the
   exact requested claim paths (here `[org.iso.18013.5.1, sub]`).
3. The OIDF conformance suite generates its mock mDL from the **ISO 18013-5 mDL document
   type** (multipaz `TestAppUtils`): it only creates elements defined in the mDL doctype
   (`given_name`, `family_name`, `birth_date`, `document_number`, ...). `sub` and `username`
   are **not ISO mDL elements**, so the mock mDL can never contain them.

Net effect: **no standard ISO 18013-5 mDL can be accepted** for login, because real mDLs do
not carry `sub`/`username`. The rejection happens before any cryptographic/trust
verification of the credential.

#### Expected behavior

The verifier should be able to authenticate a user with a standard ISO 18013-5 mDL (which
the conformance suite's mock mDL is), i.e. the identity claims used to resolve the Keycloak
user must be **configurable per credential/profile** instead of being hard-coded to
`sub`/`username`.

#### Suggested fix

- Add optional identity-claim configuration to the credential requirement (e.g.
  `subjectClaim` / `usernameClaim` in the profile JSON), defaulting to `sub` / `username`
  for backwards compatibility.
- Use the configured claims in `OID4VPProfileConfig.validatePrimaryCredential()` and in
  `OID4VPAuthenticator.recoverUserFromClaims()` (currently reads hard-coded `sub` and
  `username`).
- For the mDoc conformance profile, map identity from a standard mDL claim (e.g.
  `org.iso.18013.5.1/document_number` → username) and request only claims the presented
  credential can contain.
- Add regression tests (e.g. `OID4VPUserAuthEndpointHAIPTest`, `MdocVerificationTest`) with
  a standard mDL fixture that carries no `sub`/`username`.

#### Impact

Without this fix, the verifier cannot complete authentication with any standard ISO mDL,
the OID4VP happy-flow conformance module cannot pass for `credential_format=iso_mdl`.

Vitalisn4:
#### Summary
 
The primary credential was hard-coded to require `sub` and `username` claims to
recover the authenticating user. Standard ISO 18013-5 mDLs carry neither their
identity attributes live in namespaces (e.g. `org.iso.18013.5.1/document_number`)
so every valid mDoc presentation was rejected at DCQL pre-validation before the
signature was even checked.
 
#### Change
 
- Add optional `subjectClaim` / `usernameClaim` to `CredentialRequirement`,
  defaulting to `sub` / `username` for full backward compatibility.
- Validate the configured identity claims in `OID4VPProfileConfig` (blank guard,
  namespaced and bare forms supported, old error message preserved for defaults).
- Use the configured claims for user recovery in `OID4VPAuthenticator`.
- mDoc profiles can now map identity to `org.iso.18013.5.1/document_number`.
 
#### Testing
 
- 6 new regression tests (config validation, DCQL pre-validation, end-to-end auth
  with an mDoc containing only standard mDL claims).
- Full suite: 395 tests, 0 failures.
- OIDF conformance happy-flow passes end-to-end on both HAIP and plain_vp.

Closes https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/153

### DIFF
--- a/docs/modules/ROOT/pages/german-wallet-pid-verification-configuration.adoc
+++ b/docs/modules/ROOT/pages/german-wallet-pid-verification-configuration.adoc
@@ -181,55 +181,9 @@ Use a single-credential profile when only a self-issued primary credential is re
 
 == Primary credential identity claim configuration
 
-A primary credential must carry at least one claim that identifies the Keycloak user.
-Two optional properties control which claims are used for user resolution:
+When a primary mDoc credential identifies the Keycloak user, use a namespaced `subjectClaim` (e.g. `eu.europa.ec.eudi.pid.1/personal_administrative_number`) rather than a bare element like `document_number`, which may not be globally unique across issuers.
 
-`subjectClaim`::
-  The claim used to resolve the Keycloak user by *user id*.
-  Defaults to `sub`.
-
-`usernameClaim`::
-  An optional fallback claim used to resolve the Keycloak user by *username* when `subjectClaim` does not match a Keycloak user id.
-  Defaults to `username`.
-  Omit or set to blank to skip the username fallback.
-
-For mDoc credentials, claims are namespaced using the `namespace/element` format defined by ISO 18013-5.
-For example, a PID mDoc identifies a person by:
-
-[source,text]
-----
-namespace: eu.europa.ec.eudi.pid.1
-element:   given_name
-Namespaced form: eu.europa.ec.eudi.pid.1/given_name
-----
-
-When configuring a primary mDoc credential for authentication, set `subjectClaim` to a namespaced claim that uniquely identifies the user (e.g. `eu.europa.ec.eudi.pid.1/personal_administrative_number`).
-Do *not* set `subjectClaim` to a bare element like `document_number` unless it is globally unique across issuers.
-
-The following example shows the identity claim configuration for a primary mDoc credential.
-The `trust` configuration is omitted for brevity — refer to the x5c trust or EUDI PID trust-list sections above.
-
-[source,json]
-----
-{
-  "id": "main",
-  "role": "primary",
-  "format": "mso_mdoc",
-  "credentialTypes": [
-    "eu.europa.ec.eudi.pid.1"
-  ],
-  "subjectClaim": "eu.europa.ec.eudi.pid.1/personal_administrative_number",
-  "usernameClaim": "eu.europa.ec.eudi.pid.1/given_name",
-  "claims": [
-    "eu.europa.ec.eudi.pid.1/personal_administrative_number",
-    "eu.europa.ec.eudi.pid.1/given_name",
-    "eu.europa.ec.eudi.pid.1/family_name"
-  ],
-  "trust": [ ... ]
-}
-----
-
-For SD-JWT VC credentials, use standard claim names such as `sub` and `username` — no namespace is needed.
+For full details on `subjectClaim`, `usernameClaim`, and the namespace/element format, see xref:oid4vp-deployment-configuration.adoc#oid4vp-deployment-configuration[OpenID4VP deployment configuration reference — Configuring primary credential identity claims].
 
 == Authentication flow
 
--- a/docs/modules/ROOT/pages/german-wallet-pid-verification-configuration.adoc
+++ b/docs/modules/ROOT/pages/german-wallet-pid-verification-configuration.adoc
@@ -179,6 +179,58 @@ Use a single-credential profile when only a self-issued primary credential is re
 ]
 ----
 
+== Primary credential identity claim configuration
+
+A primary credential must carry at least one claim that identifies the Keycloak user.
+Two optional properties control which claims are used for user resolution:
+
+`subjectClaim`::
+  The claim used to resolve the Keycloak user by *user id*.
+  Defaults to `sub`.
+
+`usernameClaim`::
+  An optional fallback claim used to resolve the Keycloak user by *username* when `subjectClaim` does not match a Keycloak user id.
+  Defaults to `username`.
+  Omit or set to blank to skip the username fallback.
+
+For mDoc credentials, claims are namespaced using the `namespace/element` format defined by ISO 18013-5.
+For example, a PID mDoc identifies a person by:
+
+[source,text]
+----
+namespace: eu.europa.ec.eudi.pid.1
+element:   given_name
+Namespaced form: eu.europa.ec.eudi.pid.1/given_name
+----
+
+When configuring a primary mDoc credential for authentication, set `subjectClaim` to a namespaced claim that uniquely identifies the user (e.g. `eu.europa.ec.eudi.pid.1/personal_administrative_number`).
+Do *not* set `subjectClaim` to a bare element like `document_number` unless it is globally unique across issuers.
+
+The following example shows the identity claim configuration for a primary mDoc credential.
+The `trust` configuration is omitted for brevity — refer to the x5c trust or EUDI PID trust-list sections above.
+
+[source,json]
+----
+{
+  "id": "main",
+  "role": "primary",
+  "format": "mso_mdoc",
+  "credentialTypes": [
+    "eu.europa.ec.eudi.pid.1"
+  ],
+  "subjectClaim": "eu.europa.ec.eudi.pid.1/personal_administrative_number",
+  "usernameClaim": "eu.europa.ec.eudi.pid.1/given_name",
+  "claims": [
+    "eu.europa.ec.eudi.pid.1/personal_administrative_number",
+    "eu.europa.ec.eudi.pid.1/given_name",
+    "eu.europa.ec.eudi.pid.1/family_name"
+  ],
+  "trust": [ ... ]
+}
+----
+
+For SD-JWT VC credentials, use standard claim names such as `sub` and `username` — no namespace is needed.
+
 == Authentication flow
 
 At runtime:
--- a/docs/modules/ROOT/pages/oid4vp-user-authentication.adoc
+++ b/docs/modules/ROOT/pages/oid4vp-user-authentication.adoc
@@ -110,31 +110,9 @@ For detailed deployment options (`x509_hash`, `direct_post.jwt`, injected certif
 
 === Configuring primary credential identity claims
 
-When the primary credential is used to identify the Keycloak user, two optional properties control which claims are used for user resolution.
+When a primary credential is used to identify the Keycloak user, use `subjectClaim` and `usernameClaim` to control which claims are read from the presented credential.
 
-`subjectClaim`::
-  The claim used to resolve the Keycloak user by *user id*.
-  Defaults to `sub`.
-
-`usernameClaim`::
-  An optional fallback claim used to resolve the Keycloak user by *username* when `subjectClaim` does not match a Keycloak user id.
-  Defaults to `username`.
-  Omit or set to blank to skip the username fallback.
-
-For mDoc credentials, claims are namespaced using the `namespace/element` format defined by ISO 18013-5.
-For example, a PID mDoc identifies a person by:
-
-[source,text]
-----
-namespace: eu.europa.ec.eudi.pid.1
-element:   given_name
-Namespaced form: eu.europa.ec.eudi.pid.1/given_name
-----
-
-When setting `subjectClaim` for an mDoc primary credential, use a namespaced claim that uniquely identifies the user (e.g. `eu.europa.ec.eudi.pid.1/personal_administrative_number`).
-Do *not* set `subjectClaim` to a bare element like `document_number` unless it is globally unique across issuers.
-
-For SD-JWT VC credentials, use standard claim names such as `sub` and `username` — no namespace is needed.
+For the full configuration reference — including the mDoc namespace/element format, defaults, and examples — see link:oid4vp-deployment-configuration.adoc[OpenID4VP deployment configuration reference — Configuring primary credential identity claims].
 
 === Technical Configuration (Performance & Security)
 
--- a/docs/modules/ROOT/pages/oid4vp-user-authentication.adoc
+++ b/docs/modules/ROOT/pages/oid4vp-user-authentication.adoc
@@ -108,6 +108,34 @@ The single SdJwt authenticator in the flow is pre-configured with sensible defau
 These policy settings are deployed through the same authenticator configuration surface.
 For detailed deployment options (`x509_hash`, `direct_post.jwt`, injected certificates, and infrastructure-as-code mapping guidance), see link:oid4vp-deployment-configuration.adoc[OpenID4VP deployment configuration reference].
 
+=== Configuring primary credential identity claims
+
+When the primary credential is used to identify the Keycloak user, two optional properties control which claims are used for user resolution.
+
+`subjectClaim`::
+  The claim used to resolve the Keycloak user by *user id*.
+  Defaults to `sub`.
+
+`usernameClaim`::
+  An optional fallback claim used to resolve the Keycloak user by *username* when `subjectClaim` does not match a Keycloak user id.
+  Defaults to `username`.
+  Omit or set to blank to skip the username fallback.
+
+For mDoc credentials, claims are namespaced using the `namespace/element` format defined by ISO 18013-5.
+For example, a PID mDoc identifies a person by:
+
+[source,text]
+----
+namespace: eu.europa.ec.eudi.pid.1
+element:   given_name
+Namespaced form: eu.europa.ec.eudi.pid.1/given_name
+----
+
+When setting `subjectClaim` for an mDoc primary credential, use a namespaced claim that uniquely identifies the user (e.g. `eu.europa.ec.eudi.pid.1/personal_administrative_number`).
+Do *not* set `subjectClaim` to a bare element like `document_number` unless it is globally unique across issuers.
+
+For SD-JWT VC credentials, use standard claim names such as `sub` and `username` — no namespace is needed.
+
 === Technical Configuration (Performance & Security)
 
 For details on certificate caching, lifespan, and other performance/security configurations, see link:non-functional-requirements.adoc[Non-Functional Requirements & Technical Configuration].
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
@@ -22,7 +22,6 @@ import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import org.jboss.logging.Logger;
-import org.keycloak.OAuth2Constants;
 import org.keycloak.authentication.AuthenticationFlowContext;
 import org.keycloak.authentication.AuthenticationFlowError;
 import org.keycloak.authentication.Authenticator;
@@ -33,7 +32,6 @@ import org.keycloak.models.KeycloakSession;
 import org.keycloak.models.RealmModel;
 import org.keycloak.models.UserModel;
 import org.keycloak.models.UserProvider;
-import org.keycloak.representations.JsonWebToken;
 import org.keycloak.representations.idm.OAuth2ErrorRepresentation;
 import org.keycloak.sessions.AuthenticationSessionModel;
 import org.keycloak.util.JsonSerialization;
@@ -255,8 +253,8 @@ public class OID4VPAuthenticator implements Authenticator {
     private UserModel recoverUserFromClaims(
             Context ctx, CredentialRequirement primaryCredentialReq, JsonNode primaryClaims) {
         CredentialVerifier verifier = ctx.credentialVerifiers().get(primaryCredentialReq.getId());
-        String subject = verifier.readClaim(primaryClaims, JsonWebToken.SUBJECT);
-        String username = verifier.readClaim(primaryClaims, OAuth2Constants.USERNAME);
+        String subject = verifier.readClaim(primaryClaims, primaryCredentialReq.getSubjectClaim());
+        String username = verifier.readClaim(primaryClaims, primaryCredentialReq.getUsernameClaim());
         logger.debugf("Attempting user recovery with subject '%s' and username '%s'", subject, username);
 
         KeycloakSession session = ctx.authenticationFlowContext().getSession();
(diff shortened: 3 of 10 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
```diff
@@ -255,8 +253,8 @@ private CredentialRequirement getPresentedPrimaryCredential(Context ctx) {
     private UserModel recoverUserFromClaims(
             Context ctx, CredentialRequirement primaryCredentialReq, JsonNode primaryClaims) {
         CredentialVerifier verifier = ctx.credentialVerifiers().get(primaryCredentialReq.getId());
-        String subject = verifier.readClaim(primaryClaims, JsonWebToken.SUBJECT);
-        String username = verifier.readClaim(primaryClaims, OAuth2Constants.USERNAME);
+        String subject = verifier.readClaim(primaryClaims, primaryCredentialReq.getSubjectClaim());
```
forkimenjeckayang, line 256: `document_number` falls through to a realm-wide `getUserByUsername()` lookup, while the verified credential issuer is discarded. If this profile accepts more than one issuer, issuer B can authenticate as issuer A's user by presenting the same document number.

 OID4VP Final §14.4 requires the identity claim to be combined with the Credential Issuer identifier. Resolve a persisted `(issuer, configured-claim-value)` association using a stable trust-framework or configured issuer ID—not a rotating leaf-certificate fingerprint—or explicitly enforce a single issuer. Please add a two-issuer, same-document-number regression test.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/OID4VPProfileConfig.java
```diff
@@ -398,12 +399,39 @@ private static void validatePrimaryCredential(AuthenticationProfile profile, Cre
...
+     * Whether {@code refs} contains the configured identity claim. An un-namespaced configured claim
+     * matches by name under any namespace (so mDoc profiles requesting e.g.
+     * {@code org.iso.18013.5.1/sub} still satisfy the default {@code sub}); a namespaced configured
+     * claim must match the exact namespace/name pair.
+     */
+    private static boolean containsIdentityClaim(List<ClaimReference> refs, ClaimReference configured) {
+        if (configured.isNamespaced()) {
+            return refs.contains(configured);
         }
+        return refs.stream().anyMatch(ref -> configured.name().equals(ref.name()));
```
forkimenjeckayang, line 451: For `mso_mdoc`, accepting a bare configured identity claim is ambiguous. This validation accepts `subjectClaim: "document_number"` when the DCQL request contains `org.iso.18013.5.1/document_number`, but `MdocCredentialVerifier.readClaim()` later searches every namespace and returns the first `document_number` it finds. An issuer can carry an extension namespace with the same element name, so user recovery may use a different claim than the requested standard mDL path.

Please require `subjectClaim` and `usernameClaim` to be namespace-qualified when `format` is `mso_mdoc`, and match them exactly to a requested claim path. Add a negative configuration test for a bare mDoc identity selector and a runtime test with duplicate element identifiers in two namespaces.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/CredentialRequirement.java
```diff
@@ -34,6 +36,20 @@ public class CredentialRequirement {
...
+    @JsonProperty("subjectClaim")
+    private String subjectClaim = JsonWebToken.SUBJECT;
+
+    /**
+     * Claim (optionally namespaced, e.g. {@code "org.iso.18013.5.1/document_number"}) carrying the
+     * credential subject used to resolve the Keycloak user by username. Defaults to {@code username}.
+     */
+    @JsonProperty("usernameClaim")
+    private String usernameClaim = OAuth2Constants.USERNAME;
+
```
forkimenjeckayang, line 55: Please document `subjectClaim` and `usernameClaim` beside the existing authentication-profile JSON. Include defaults, the `namespace/element` form required for mDoc, and a complete mDL primary example. The example should describe the issuer-scoped identity association required for authentication, rather than suggesting that a bare `document_number` is a realm-wide username.

forkimenjeckayang (changes requested):
A few comments 
Please check

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
IngridPuppet: I don't understand this comment. Please could you explain in simpler terms what is the issue? The change simply retains the previous behavior with configurability. Maybe a focused ticket should be created.

For SD-JWT, we have some mechanism to enforce that the issuer claim is that of Keycloak. mDocs do not seem to embed a comparable issuer claim on the other hand; the only and real trust remains the certificate chain validation.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/CredentialRequirement.java
IngridPuppet: Could you clarify in which file lives the "existing authentication-profile JSON"? Also, there is no specific focus on mDL; it's just for the conformance test. If mDoc in general is already documented, there is no need to provide a complete "mDL primary example" in my opinion.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/OID4VPProfileConfig.java
```diff
@@ -398,12 +399,39 @@ private static void validatePrimaryCredential(AuthenticationProfile profile, Cre
...
+        ClaimReference subjectRef = ClaimReference.parse(credential.getSubjectClaim());
+        ClaimReference usernameRef = ClaimReference.parse(credential.getUsernameClaim());
+        boolean hasSubject = containsIdentityClaim(primaryRefs, subjectRef);
+        boolean hasUsername = containsIdentityClaim(primaryRefs, usernameRef);
         if (!hasSubject || !hasUsername) {
-            throw new IllegalStateException("OpenID4VP primary credential must request sub and username: "
-                    + profile.getId() + "/" + credential.getId());
+            String message = isDefaultIdentityClaims(credential)
+                    ? "OpenID4VP primary credential must request sub and username"
+                    : "OpenID4VP primary credential must request identity claims " + subjectRef + " and " + usernameRef;
```
IngridPuppet: The message can just be: 

```java
"OpenID4VP primary credential must request identity claims " + subjectRef + " and " + usernameRef
```

By default, that would print sub and username. It is fine.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/OID4VPProfileConfig.java
```diff
@@ -398,12 +399,39 @@ private static void validatePrimaryCredential(AuthenticationProfile profile, Cre
...
+            throw new IllegalStateException(message + ": " + profile.getId() + "/" + credential.getId());
+        }
+    }
+
+    private static boolean isDefaultIdentityClaims(CredentialRequirement credential) {
+        return JsonWebToken.SUBJECT.equals(credential.getSubjectClaim())
+                && OAuth2Constants.USERNAME.equals(credential.getUsernameClaim());
+    }
```
IngridPuppet: This is not needed IMO.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapabilityTest.java
```diff
@@ -61,6 +61,31 @@ void validatesIsoSpecSample() {
...
+    void rejectsStandardIsoMdlWhenSubClaimIsRequested() {
+        // The pre-fix behavior: requesting the sub claim path rejects every standard ISO mDL.
+        String token = readResource("/mdoc/spec-sample.txt");
+        Credential credential =
+                credentialWithDocTypeAndClaims("org.iso.18013.5.1.mDL", claim("sub", "org.iso.18013.5.1", "sub"));
+
+        VerificationException error =
+                assertThrows(VerificationException.class, () -> capability.validatePresentation(credential, token));
+        assertEquals("Presented mDoc does not satisfy DCQL claim path: [org.iso.18013.5.1, sub]", error.getMessage());
+    }
```
IngridPuppet: Does this test anything contributed by this PR? These scenarios look already covered by `validatesMatchingDocTypeAndRequestedClaims` and `rejectsMissingRequestedClaim`.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/OID4VPProfileConfigTest.java
```diff
@@ -309,6 +309,85 @@ void shouldRejectCredentialIdentityPrimaryWithoutSubAndUsername() {
...
+                        "usernameClaim": "family_name"
+                      }
+                    ]
+                  }
+                ]
+                """));
+
+        IllegalStateException error = assertThrows(IllegalStateException.class, () -> new OID4VPProfileConfig(config));
+        assertEquals(
+                "OpenID4VP primary credential must request identity claims given_name and family_name: login/primary",
```
IngridPuppet: I think it would clearer if the error message specifies what is the subject claim and what is the username claim.
```suggestion
                "OpenID4VP primary credential must request identity claims subjectClaim='given_name' and usernameClaim='family_name': login/primary",
```

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointTest.java
```diff
@@ -380,6 +380,26 @@ public void shouldAuthenticateSuccessfully_WithMdocPrimaryCredential() throws Ex
...
 
+    @Test
+    public void shouldAuthenticateSuccessfully_WithMdocPrimaryCredential_MdlIdentityClaim() throws Exception {
+        // Regression for issue 001: a standard ISO 18013-5 mDL carries no sub/username claims, so
+        // identity must be derivable from a configured standard mDL claim (document_number). The
+        // presented mDoc below deliberately contains no sub/username claims.
+        withAuthenticationProfile(
+                AuthenticationProfileSamples.mdocPrimaryWithMdlIdentity(), (apiFlow, requestObject) -> {
+                    Map<String, Object> mdocClaims =
+                            Map.of(MdocBaseTest.NAMESPACE, Map.of("document_number", TEST_USER, "given_name", "Alice"));
```
IngridPuppet, line 396: In other tests, we have:

```
Map.of(JsonWebToken.SUBJECT, TEST_USER_ID, OAuth2Constants.USERNAME, TEST_USER)
```

So `sub` is not the same thing as `username`: `TEST_USER_ID` vs `TEST_USER`. 

Here, it looks like the value of `document_number` is used for both. Could you help me understand how that even pass authentication?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/OID4VPProfileConfig.java
```diff
@@ -398,12 +399,39 @@ private static void validatePrimaryCredential(AuthenticationProfile profile, Cre
             }
             return;
         }
+        if (StringUtil.isBlank(credential.getSubjectClaim()) || StringUtil.isBlank(credential.getUsernameClaim())) {
+            throw new IllegalStateException(
+                    "OpenID4VP primary credential subjectClaim and usernameClaim must not be blank: " + profile.getId()
+                            + "/" + credential.getId());
+        }
```
IngridPuppet: Could you make `usernameClaim` optional? Actually, it is just a temporary workaround. It is fine if subjectClaim alone enables recovering the user.

IngridPuppet (changes requested):
I added a few more comments. Please could you check?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
forkimenjeckayang: You’re right the PR only makes the existing user lookup configurable, so my concern is outside its scope.
In simple terms, certificate validation confirms that an mDoc comes from a trusted issuer, but the user lookup then uses only `document_number` and does not retain which issuer provided it. OID4VP Sec 14.4 recommends combining both values. This broader identity-mapping concern should be handled in a separate ticket and should not block this PR.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/CredentialRequirement.java
forkimenjeckayang: The existing profile JSON I meant is in `docs/modules/ROOT/pages/german-wallet-pid-verification-configuration.adoc`.  My main request is only to document the new subjectClaim and usernameClaim properties, their defaults, and the namespace/element format required for mDoc. This could be added to `docs/modules/ROOT/pages/oid4vp-user-authentication.adoc.`

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointTest.java
Vitalisn4: When the plugin tries to log the user in, it first tries to find the user by ID using the  subjectClaim  value. That fails because  987654321  (the document number) is not a Keycloak user ID (which is a UUID). So it falls back to finding the user by username using the  usernameClaim  value. That succeeds because we created a Keycloak user with that document number as the username. So even though both values are the same, the second lookup is the one that actually finds the user.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/dcql/MdocDcqlCredentialCapabilityTest.java
Vitalisn4: That's right. The test was checking that requesting a sub-claim from a standard mDL fails, but that is already covered by the existing `rejectsMissingRequestedClaim` test. So I removed the duplicate. The other test in that file (`validatesStandardIsoMdlWithoutSubOrUsernameClaims`) was kept because it tests the actual fix username.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/CredentialRequirement.java
IngridPuppet: Thank you for the clarification. I believe there might be some confusion between mDoc and mDL. Because it targets a PID, which is an mDoc, I agree the new attributes should be documented in `docs/modules/ROOT/pages/german-wallet-pid-verification-configuration.adoc`. However, a PID is not an mDL; no focus should be placed on mDL.

> My main request is only to document the new subjectClaim and usernameClaim properties, their defaults, and the namespace/element format required for mDoc. This could be added to `docs/modules/ROOT/pages/oid4vp-user-authentication.adoc.`

Certainly. Thank you. I hope it is clear for @Vitalisn4 how to update the documentation to address this point.

Vitalisn4: > Thank you for the clarification. I believe there might be some confusion between mDoc and mDL. Because it targets a PID, which is an mDoc, I agree the new attributes should be documented in `docs/modules/ROOT/pages/german-wallet-pid-verification-configuration.adoc`. However, a PID is not an mDL; no focus should be placed on mDL.
> 
> > My main request is only to document the new subjectClaim and usernameClaim properties, their defaults, and the namespace/element format required for mDoc. This could be added to `docs/modules/ROOT/pages/oid4vp-user-authentication.adoc.`
> 
> Certainly. Thank you. I hope it is clear for @Vitalisn4 how to update the documentation to address this point.

Yes, it's clear for me now and will work on that

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointTest.java
IngridPuppet: Is TEST_USER worth 987654321? Where does that number come from?

docs/modules/ROOT/pages/oid4vp-user-authentication.adoc
```diff
@@ -108,6 +108,34 @@ The single SdJwt authenticator in the flow is pre-configured with sensible defau
...
+----
+namespace: eu.europa.ec.eudi.pid.1
+element:   given_name
+Namespaced form: eu.europa.ec.eudi.pid.1/given_name
+----
+
+When setting `subjectClaim` for an mDoc primary credential, use a namespaced claim that uniquely identifies the user (e.g. `eu.europa.ec.eudi.pid.1/personal_administrative_number`).
+Do *not* set `subjectClaim` to a bare element like `document_number` unless it is globally unique across issuers.
+
+For SD-JWT VC credentials, use standard claim names such as `sub` and `username` — no namespace is needed.
```
IngridPuppet: This information does belong to this documentation page, which aims to focus on the end to end flow, not too technical configuration. Move it to `oid4vp-deployment-configuration` instead.

docs/modules/ROOT/pages/german-wallet-pid-verification-configuration.adoc
```diff
@@ -179,6 +179,58 @@ Use a single-credential profile when only a self-issued primary credential is re
...
+  "claims": [
+    "eu.europa.ec.eudi.pid.1/personal_administrative_number",
+    "eu.europa.ec.eudi.pid.1/given_name",
+    "eu.europa.ec.eudi.pid.1/family_name"
+  ],
+  "trust": [ ... ]
+}
+----
+
+For SD-JWT VC credentials, use standard claim names such as `sub` and `username` — no namespace is needed.
```
IngridPuppet: The exact information was added to another page. Refer to it and only document the special additional consideration if any.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/profile/OID4VPProfileConfigTest.java
```diff
@@ -382,10 +387,63 @@ void shouldRejectCredentialIdentityPrimaryWithBlankConfiguredIdentityClaim() {
                 ]
                 """));
 
+        assertDoesNotThrow(() -> new OID4VPProfileConfig(config));
+    }
+
+    @Test
+    void shouldRejectMdocPrimaryCredentialWithBareSubjectClaim() {
+        // forkimenjeckayang comment: bare subjectClaim is ambiguous for mso_mdoc because
+        // MdocCredentialVerifier.readClaim() searches every namespace and returns the first match.
```
IngridPuppet: ???

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/OID4VPUserAuthEndpointTest.java
Vitalisn4: The `987654321` came from the conformance test platform. When I ran the OID4VP happy-flow test, the platform generated the test log with document_number set to 987654321. I then created a Keycloak user with that number as the username to test the authentication flow. The unit test in the code uses TEST_USER, which is "test-user" (defined in BaseKeycloakTest.java), not 987654321.

IngridPuppet: I get it. Thank you for the clarification.

IngridPuppet (approved):
My comments have been addressed.

forkimenjeckayang (approved):

PR #158, opened 2026-08-18
Implement nested OpenID4VP flow for presentation during issuance on v1.0

### LINKED ISSUE #145: Implement nested OpenID4VP flow for presentation during issuance on v1.0
We would like to support presentation during issuance on OpenID4VP 1.0 according to this flow: https://bmi.usercontent.opencode.de/eudi-wallet/wallet-development-documentation-public/latest/architecture-concept/03-data-flows/31-eaa-issuance.html#presentation-during-issuance

Also discussed here: https://github.com/openid/OpenID4VCI/issues/730

ETA: 3 days

IngridPuppet:
Closes #145

### DIFF
--- a/docs/modules/ROOT/pages/presentation-during-issuance-configuration.adoc
+++ b/docs/modules/ROOT/pages/presentation-during-issuance-configuration.adoc
@@ -1,0 +1,111 @@
+= Configuring presentation during credential issuance
+
+This page describes how to require a verified OpenID4VP presentation before Keycloak issues a
+specific OpenID4VCI credential. Presentation during issuance is configured at two levels:
+
+* the realm feature flag enables the capability;
+* the credential configuration declares which presentation mode or modes are required.
+
+The plugin supports two presentation-during-issuance modes:
+
+* `interactive_authorization`, where the wallet presents through the OID4VCI authorization challenge endpoint;
+* `nested_oid4vp_flow`, where the presentation is nested inside the normal OIDC authorization flow.
+
+The gate applies only to credential configurations that explicitly require a presentation. Other
+credentials continue to use their normal issuance flow.
+
+== Prerequisites
+
+* An OpenID4VP authentication flow and profile configured for the presentation.
+* An issuing client configured to request the credential.
+
+== Enable the realm capability
+
+Set the following realm attribute:
+
+[source,json]
+----
+{
+  "attributes": {
+    "oid4vci.presentation_during_issuance": "true"
+  }
+}
+----
+
+The same attribute can be set through the Admin REST API or by including it in a realm import.
+The value is a string and must be `"true"`.
+
+When enabled, the realm exposes the OID4VCI authorization challenge endpoint and permits the
+issuance flows described below. The realm flag alone does not gate any credential.
+
+== Configure a gated credential
+
+Update the configuration of the credential with these relevant attributes:
+
+[source,text]
+----
+vc.requires_presentation: interactive_authorization
+vc.presentation_profile_id: dual-german-pid
+----
+
+`vc.requires_presentation` declares the accepted presentation-during-issuance modes. Omitting the
+attribute means that the credential is not gated. A present but empty value accepts any supported
+mode.
+
+`vc.presentation_profile_id` identifies the OpenID4VP authentication profile that validates the
+required presentation. The profile defines the requested credential, claims, holder binding, and
+trust policy. If the profile uses session identity, start with an authorization-code credential offer
+that contains `issuer_state`; Keycloak uses that state to resolve the offer's target user before
+validating the presentation.
+
+== Select a presentation mode
+
+`vc.requires_presentation` accepts one or more comma-separated values:
+
+* `interactive_authorization`
+* `nested_oid4vp_flow`
+
+For example, to accept either mode:
+
+[source,text]
+----
+vc.requires_presentation: interactive_authorization,nested_oid4vp_flow
+----
+
+=== Interactive authorization
+
+Use `interactive_authorization` when the wallet starts at the OID4VCI authorization challenge
+endpoint. The challenge returns an OpenID4VP request. The wallet presents the required credential,
+and the resulting authorization code can then be used to request the issued credential.
+
+The credential endpoint rejects issuance if the access-token session does not contain a verified
+presentation obtained through an accepted mode.
+
+=== Nested OID4VP flow
+
+Use `nested_oid4vp_flow` when the presentation is nested inside the normal OIDC authorization
+request. When an authorization request targets a credential configuration with this mode, the
+normal username/password login view is replaced by a same-device OpenID4VP presentation. After
+verification, Keycloak resumes the OIDC authorization flow and issues the authorization code.
+
+== Runtime behavior
+
+For a gated credential, the expected flow is:
+
+. The client requests the gated credential configuration.
+. Keycloak creates or resumes the required authorization flow.
+. The wallet presents the credential described by the configured OpenID4VP profile.
+. Keycloak verifies the presentation, including its format, claims, holder binding, and trust policy.
+. Keycloak records the verified presentation mode in the authenticated session.
+. The wallet requests the credential from the OID4VCI credential endpoint.
+. Keycloak issues the credential only when the recorded mode is accepted by `vc.requires_presentation`.
+
+Issuance is rejected when the presentation is missing, invalid, or was obtained through a mode not
+listed in the credential configuration. The pre-authorized-code path cannot bypass this requirement
+for a gated credential.
+
+== References
+
+* https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html[OpenID4VCI specification]
+* xref:oid4vp-user-authentication.adoc[Authenticate users with verifiable credentials]
+* xref:oid4vp-deployment-configuration.adoc[OpenID4VP deployment configuration]
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/PresentationVerifiedNote.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/PresentationVerifiedNote.java
@@ -1,0 +1,30 @@
+package io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.keycloak.util.JsonSerialization;
+
+/**
+ * Value stored in the user-session note that records a successfully verified presentation during
+ * issuance. It binds the {@link PresentationDuringIssuanceMode} through which the presentation was
+ * obtained to the concrete OpenID4VP authentication profile that was used, so the issuance gate can
+ * verify the proof against the specific credential configuration.
+ *
+ * @param mode the {@link PresentationDuringIssuanceMode} value
+ * @param profileId the OpenID4VP authentication profile id used
+ */
+public record PresentationVerifiedNote(
+        @JsonProperty("mode") String mode,
+        @JsonProperty("profileId") String profileId) {
+
+    public static PresentationVerifiedNote of(PresentationDuringIssuanceMode mode, String profileId) {
+        return new PresentationVerifiedNote(mode.getValue(), profileId);
+    }
+
+    public String toJson() {
+        return JsonSerialization.valueAsString(this);
+    }
+
+    public static PresentationVerifiedNote fromJson(String json) {
+        return JsonSerialization.valueFromString(json, PresentationVerifiedNote.class);
+    }
+}
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
@@ -9,7 +9,7 @@ import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.dcql.DcqlCredentialC
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.dcql.DcqlCredentialCapability;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.dcql.DcqlResponseCredentialSelector;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model.PresentationDuringIssuanceMode;
-import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model.ResponseMode;
+import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model.PresentationVerifiedNote;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model.ResponseObject;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model.dcql.Credential;
 import io.github.adorsysgis.keycloak.protocol.oid4vc.oid4vp.model.dcql.DcqlQuery;
@@ -54,7 +54,12 @@ public class AuthorizationResponseService {
 
     private static final Logger logger = Logger.getLogger(AuthorizationResponseService.class);
 
-    public static final String PARENT_AUTH_SESSION_ID = "parent_auth_session_id";
+    /**
+     * Client-session note carrying the id of the parent OIDC authentication session that this
+     * OpenID4VP authorization is bound to. Used by {@code OID4VPLoginActionsService} to verify that
+     * a redeemed authorization code belongs to the OIDC session resuming the login.
+     */
+    public static final String PARENT_AUTH_SESSION_ID = "oid4vp.parent_auth_session_id";
 
     private final KeycloakSession session;
     private final DcqlCredentialCapabilities dcqlCapabilities;
@@ -268,31 +273,9 @@ public class AuthorizationResponseService {
      */
     private String produceAuthorizationCode(
             AuthenticatedClientSessionModel clientSession, AuthorizationContext authContext) {
-        // Decorate client session with contextual notes
-
-        if (authContext.getParentAuthSessionId() != null) {
-            clientSession.setNote(PARENT_AUTH_SESSION_ID, authContext.getParentAuthSessionId());
-        }
-
-        clientSession.setNote(
-                OIDCLoginProtocol.ISSUER,
-                Urls.realmIssuer(
-                        session.getContext().getUri().getBaseUri(),
-                        session.getContext().getRealm().getName()));
-
-        clientSession.setNote(PARAM_LOGIN_METHOD, LOGIN_METHOD_OID4VP);
-
-        // Mark the user session as presentation-verified through the interactive authorization mode
-        // (Authorization Challenge Endpoint / OID4VCI §6). Presence verifies that the authorization code
-        // was obtained via a Verifiable Presentation, and the value annotates the exact mode for the
-        // issuance gate. This only applies when the response_mode is explicitly interactive authorization.
-        ResponseMode responseMode = authContext.getRequestObject().getResponseMode();
-        if (responseMode.isInteractiveAuthorization()) {
-            UserSessionModel userSession = clientSession.getUserSession();
-            userSession.setNote(
-                    OpenId4VpConstants.PRESENTATION_VERIFIED_NOTE,
-                    PresentationDuringIssuanceMode.INTERACTIVE_AUTHORIZATION.getValue());
-        }
+        // Decorate sessions with contextual notes
+        decorateClientSession(clientSession, authContext);
+        decorateUserSession(clientSession.getUserSession(), authContext);
 
         // Gather code data and generate authorization code
 
@@ -308,6 +291,45 @@ public class AuthorizationResponseService {
                 clientSession.getUserSession().getId());
 
         return OAuth2CodeParser.persistCode(session, clientSession, codeData);
+    }
+
+    /**
+     * Decorates the client session with contextual notes for the completed authorization.
+     */
+    private void decorateClientSession(
+            AuthenticatedClientSessionModel clientSession, AuthorizationContext authContext) {
+        clientSession.setNote(PARAM_LOGIN_METHOD, LOGIN_METHOD_OID4VP);
+
+        clientSession.setNote(
+                OIDCLoginProtocol.ISSUER,
+                Urls.realmIssuer(
+                        session.getContext().getUri().getBaseUri(),
+                        session.getContext().getRealm().getName()));
+
+        if (authContext.getParentAuthSessionId() != null) {
+            clientSession.setNote(PARENT_AUTH_SESSION_ID, authContext.getParentAuthSessionId());
+        }
+    }
+
+    /**
+     * Decorates the user session with the presentation-verified marker when the authorization serves
+     * credential issuance.
+     */
+    private void decorateUserSession(UserSessionModel userSession, AuthorizationContext authContext) {
+        // Mark the user session as presentation-verified through the mode recorded on the
+        // authorization context (Authorization Challenge Endpoint / OID4VCI §6). Presence verifies
+        // that the authorization code was obtained via a Verifiable Presentation, and the value binds
+        // the exact mode and OpenID4VP authentication profile for the issuance gate. Only sessions
+        // started as presentation-during-issuance carry a mode.
+        if (authContext.getPresentationDuringIssuanceMode() != null) {
+            PresentationDuringIssuanceMode mode =
+                    PresentationDuringIssuanceMode.fromValue(authContext.getPresentationDuringIssuanceMode());
+
+            userSession.setNote(
+                    OpenId4VpConstants.PRESENTATION_VERIFIED_NOTE,
+                    PresentationVerifiedNote.of(mode, authContext.getProfileId())
+                            .toJson());
+        }
     }
 
     /**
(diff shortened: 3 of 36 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/patch/issuance/PatchedOID4VCIssuerEndpoint.java
Ogenbertrand, line 1: I noticed something in line 80 of this file.
That gate resolves the credential policy from the first authorization_detail in the access token, not from the credential actually requested in this Credential Request. If a token authorizes multiple credential configurations, this can bypass the presentation gate for a later gated credential or reject an ungated request because an earlier entry is gated. We could parse the requested credential_identifier / credential_configuration_id from requestPayload, match it against the token authorization_details, and gate that exact credential. OID4VCI 1.0 allows authorization_details to carry credential identifiers used later at the Credential Endpoint: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-6.2

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/OID4VPLoginActionsService.java
```diff
@@ -153,7 +156,15 @@ public Response oid4vpAuthLogin(
                 result.getClientSession().getUserSession().getUser());
         ClientSessionContext clientSessionCtx =
                 AuthenticationProcessor.attachSession(authSession, null, session, realm, clientConnection, event);
+
+        // Mark the user session as presentation-verified through the nested OID4VP flow, which
+        // resumes the OIDC login through this service after the same-device presentation. This
+        // mirrors `AuthorizationResponseService.produceAuthorizationCode` for presentation
+        // during issuance via interactive authorization.
         UserSessionModel freshUserSession = clientSessionCtx.getClientSession().getUserSession();
```
Ogenbertrand, line 165: The verified-presentation marker is stored only as a broad UserSession note with the mode value, so it can be reused by later issuance requests in the same SSO session regardless of credential configuration/profile. For presentation during issuance, the proof should be bound to the concrete issuance authorization/requested credential/profile.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
```diff
@@ -279,20 +282,27 @@ private String produceAuthorizationCode(
...
-        // authorization code was obtained via a Verifiable Presentation (OID4VCI Interactive
-        // Authorization). This marker is the carrier consumed at the credential endpoint.
-        clientSession.getUserSession().setNote(OpenId4VpConstants.PRESENTATION_VERIFIED_NOTE, Boolean.TRUE.toString());
+        // Mark the user session as presentation-verified through the interactive authorization mode
+        // (Authorization Challenge Endpoint / OID4VCI §6). Presence verifies that the authorization code
+        // was obtained via a Verifiable Presentation, and the value annotates the exact mode for the
+        // issuance gate. This only applies when the response_mode is explicitly interactive authorization.
+        ResponseMode responseMode = authContext.getRequestObject().getResponseMode();
+        if (responseMode.isInteractiveAuthorization()) {
+            UserSessionModel userSession = clientSession.getUserSession();
```
Ogenbertrand: Same concern as the nested flow: the interactive authorization proof is recorded only on the UserSession. Please bind the success marker to the authorization transaction/requested credential instead of allowing any later gated credential in the same session to consume it.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/presentation/GuardedCredentialScope.java
```diff
@@ -0,0 +1,79 @@
...
+    public static GuardedCredentialScope from(CredentialScopeModel credentialScope) {
+        return credentialScope == null ? null : new GuardedCredentialScope(credentialScope);
+    }
+
+    /**
+     * Whether the credential requires a Verifiable Presentation prior to issuance.
+     */
+    public boolean requiresPresentation() {
+        return getAttribute(VC_REQUIRES_PRESENTATION_ATTR) != null;
+    }
```
Ogenbertrand, line 81: This changes the old boolean-like semantics: `vc.requires_presentation=false` now means “presentation required”, and then no real mode will match.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/PresentationDuringIssuanceService.java
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * Resolves the requested credential from the server-side credential offer referenced by issuer_state.
+     */
+    private static List<CredentialScopeModel> resolveFromIssuerState(
+            AuthenticationSessionModel authSession, CredentialOfferState requestedCredentialOfferState) {
+        if (requestedCredentialOfferState == null) {
+            String issuerState = authSession.getClientNote(ISSUER_STATE_NOTE);
+            return StringUtil.isBlank(issuerState) ? null : List.of();
+        }
```
Ogenbertrand, line 128: If issuer_state is present but the referenced offer is missing/expired, this returns an empty list and effectively falls back into “no requested credential”. issuer_state is the server-side binding to the credential offer/target user, so this should fail closed with a clear invalid/expired issuer_state error instead of silently disabling the nested presentation decision.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/patch/issuance/PatchedOIDCLoginProtocol.java
```diff
@@ -0,0 +1,57 @@
...
+import org.keycloak.protocol.oidc.OIDCConfigAttributes;
+import org.keycloak.protocol.oidc.OIDCLoginProtocol;
+import org.keycloak.protocol.oidc.OIDCProviderConfig;
+import org.keycloak.sessions.AuthenticationSessionModel;
+
+/**
+ * OIDC patch for clients that send {@code dpop_jkt} without enabling DPoP.
+ *
+ * <p>See <a href="https://github.com/keycloak/keycloak/issues/51573">Keycloak issue #51573</a>.
+ */
```
Ogenbertrand, line 22: This class is documented as a DPoP workaround, but it now also forces reauthentication for nested presentation during issuance. Can you update the class documentation/name or split the responsibility so future maintainers do not miss that this is part of the issuance security flow.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/NestedPresentationDuringIssuanceTest.java
```diff
@@ -0,0 +1,441 @@
...
+    @Test
+    @DisplayName("should reject multiple openid_credential authorization details")
+    void shouldRejectMultipleCredentialAuthorizationDetails() throws Exception {
+        // Nested issuance supports exactly one credential configuration per authorization request.
+        String details = authorizationDetails(CREDENTIAL_IDENTITY_CONFIG_ID, SESSION_IDENTITY_CONFIG_ID);
+        String authEndpoint = buildAuthorizationEndpoint(null, details, null);
+
+        Connection.Response response = requestAuthorizationPage(authEndpoint);
+        assertEquals(HttpStatus.SC_BAD_REQUEST, response.statusCode());
+        assertTrue(response.body().contains("An internal server error has occurred"));
```
Ogenbertrand, line 403: This test asserts an internal-server-error-looking page for a controlled request/configuration error. Please surface multiple requested credential configurations as a proper OAuth/Keycloak error response instead of relying on a generic server error page.

Ogenbertrand (changes requested):
Hello @IngridPuppet.
Wonderful contribution, thanks for the PR.
I have a few comments for you could you check them ?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/freemarker/OID4VPUserAuthBean.java
```diff
@@ -158,16 +181,33 @@ public AuthContextBean getAuthContext() {
...
                 clientId,
                 profileId,
-                new OIDCAuthSession(authSessionId, getLoginActionUrl(), enableSameDeviceResponse),
-                codeChallengeDetails);
+                new OIDCAuthSession(authSessionId(), getLoginActionUrl(), true),
+                null,
+                subjectUserId);
+
+        return new AuthContextBean().setAuthReqLink(authContext.getAuthorizationRequest());
+    }
```
forkimenjeckayang, line 209: This needs to populate `authContextBean` before returning, as the non-PDI branch does. The template evaluates `oid4vp.authContext` once for `error` and again for `authReqLink`; because this return value is not cached, rendering one PDI page starts two different VP transactions and discards the first one.

can we assign the PDI result to `authContextBean` (including the error case) and add a test that resolves both properties or renders the template and verifies `startAuthentication()` is called exactly once.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/PresentationDuringIssuanceService.java
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * Resolves credential offer state from expected issuer state client note.
+     */
+    private static CredentialOfferState resolveCredentialOfferState(
+            KeycloakSession session, AuthenticationSessionModel authSession) {
+        String offerId = Optional.ofNullable(authSession)
+                .map(s -> s.getClientNote(ISSUER_STATE_NOTE))
+                .filter(StringUtil::isNotBlank)
+                .map(IssuerState::fromEncodedString)
```
forkimenjeckayang: can we validate and translate malformed `issuer_state` here just as `AuthorizationChallengeEndpoint.resolveCredentialOffer()` does. `IssuerState.fromEncodedString()` throws for an arbitrary non-base64 value, and this nested route currently turns that wallet/browser input into an internal server error instead of a controlled `invalid_request`/bad-request response.

Reuse a shared offer resolver for both modes and add nested-flow tests for malformed and unknown/expired `issuer_state`.

docs/modules/ROOT/pages/presentation-during-issuance-configuration.adoc
```diff
@@ -0,0 +1,111 @@
+= Configuring presentation during credential issuance
+
+This page describes how to require a verified OpenID4VP presentation before Keycloak issues a
+specific OpenID4VCI credential. Presentation during issuance is configured at two levels:
+
+* the realm feature flag enables the capability;
```
forkimenjeckayang, line 6: is this `oid4vci.presentation_during_issuance` currently a feature flag ?
the runtime nested flow and credential gate never read `oid4vci.presentation_during_issuance`; only issuer metadata does. With this attribute unset/false, a credential containing `vc.requires_presentation` is still gated and the challenge endpoint is still live, just no longer advertised.

 we can either enforce the flag consistently at every runtime entry point or document it precisely as a metadata-advertisement toggle and also add enabled and disabled integration tests so a deployment can safely predict the effect of this setting.
WDYT?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/PresentationDuringIssuanceService.java
```diff
@@ -0,0 +1,201 @@
...
+
+        if (requestedCredentials.size() > 1) {
+            String errorMessage = "Credential endpoint does not support issuing multiple credential types";
+            logger.debugf(
+                    "%s (resolved %d credential configurations under request for issuance)",
+                    errorMessage, requestedCredentials.size());
+            throw new BadRequestException(errorMessage);
+        }
+
+        return GuardedCredentialScope.from(requestedCredentials.getFirst());
```
forkimenjeckayang, line 118: could we  extract the issuer-state / `authorization_details` / scope resolution into one shared, server-side resolver and make its cardinality contract explicit. The new nested resolver rejects multiple credential configurations, whereas `AuthorizationChallengeEndpoint` and the issuer guard select the first one or this is intended

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/presentation/GuardedCredentialScope.java
```diff
@@ -0,0 +1,79 @@
...
+            return false;
+        }
+
+        String config = getAttribute(VC_REQUIRES_PRESENTATION_ATTR);
+        if (StringUtil.isBlank(config)) {
+            return true;
+        }
+
+        return Arrays.asList(config.split("[,\\s]+")).contains(mode);
+    }
```
forkimenjeckayang, line 115: Please validate `vc.requires_presentation` against the supported enum values and require a nonblank `vc.presentation_profile_id` whenever the credential is gated. At present arbitrary strings are accepted by the admin configuration and only fail after a wallet has started issuance (nested mode renders the generic error page).

Consider parsing the configured list once into `PresentationDuringIssuanceMode` values and surface a clear administrator-facing configuration error.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/NestedPresentationDuringIssuanceTest.java
forkimenjeckayang, line 1: Can we add an end-to-end test that creates an authorization-code credential offer, submits the wallet PAR, opens the returned `request_uri` in the same cookie-bearing user agent, completes the VP callback, and requests the credential?
 It verifies that `issuer_state`/credential selection and the parent authentication session survive Keycloak’s PAR/request-URI processing rather than only the direct authorization endpoint.
WDYT?

forkimenjeckayang (dismissed):
a few comments @IngridPuppet

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/OID4VPLoginActionsService.java
forkimenjeckayang: +1
There is also a concrete path here: `oid4vpAuthLogin` is shared by ordinary same-device OpenID4VP login, yet it unconditionally writes the nested-PDI marker.   The marker should be scoped to the concrete issuance authorization and requested credential/profile not a broad `UserSession` note and should only be recorded when the parent session actually initiated nested PDI. Please add coverage for ordinary same-device login, nested PDI, and later/refresh-token issuance attempts.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/patch/issuance/PatchedOID4VCIssuerEndpoint.java
IngridPuppet: > We could parse the requested credential_identifier / credential_configuration_id from requestPayload, match it against the token authorization_details, and gate that exact credential.

Obviously, this check is done in Keycloak. You can't request Credential A with an access token granted for Credential B. The access token embeds the real source of truth for credentials that can be issued. In case a credential identifier was used to request a credential, the access token also embeds the credential configuration ID that we can trust to gate.

https://github.com/keycloak/keycloak/blob/ea22b1249dec918578bf33fd65b61ba7b28941f8/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java#L996-L1023

Keycloak also enforces that only one credential configuration ID is granted by the access token. I've doubled the check for defensiveness in PatchedOID4VCIssuerEndpoint. Please confirm if this addresses your concern. Thank you.

https://github.com/keycloak/keycloak/blob/ea22b1249dec918578bf33fd65b61ba7b28941f8/services/src/main/java/org/keycloak/protocol/oid4vc/issuance/OID4VCIssuerEndpoint.java#L925-L929

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/OID4VPLoginActionsService.java
IngridPuppet: > For presentation during issuance, the proof should be bound to the concrete issuance authorization/requested credential/profile.

As implemented by Keycloak, the access token eventually issued cannot be used to request other credentials than those requested by issuance requests. This marked user session is also tied to this access token, meaning it cannot be resolved by another unrelated access token. However, I agree that the note must tell something very strict about what happened in the session because it is natural for issuance requests to share user sessions (we shouldn't fight that). It was one step to include the presentation during issuance mode. I agree it could be problematic not to also specify the authentication profile, and addressed that.

> There is also a concrete path here: oid4vpAuthLogin is shared by ordinary same-device OpenID4VP login, yet it unconditionally writes the nested-PDI marker.

The initial reasoning was: The marker means that the user session was authenticated via nested OpenID4VP, which is true for ordinary same-device OpenID4VP login. However, I understand that because the constant is tightly linked to presentation during issuance, this could be confusing and updated to only be set in that case. An existing test was updated for ordinary same-device login not to even pass the presentation gate policy.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/service/AuthorizationResponseService.java
IngridPuppet: In line with the response to https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/158#discussion_r3821021624, the note now also includes the authentication profile ID.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/presentation/GuardedCredentialScope.java
IngridPuppet: Yes, this is intended. A comment clarifies that for ungated credentials, do not configure the attribute.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/PresentationDuringIssuanceService.java
IngridPuppet: Thank you for bringing this to my attention. Actually, I should return null unconditionally so other methods to search for a credential request are tried. The issuer_state param is not OpenID4VCI-specific. It is perfectly okay to pass back some state for some other process.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/patch/issuance/PatchedOIDCLoginProtocol.java
IngridPuppet: Yes, I agree. The documentation was updated to mention the other responsibility.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/NestedPresentationDuringIssuanceTest.java
IngridPuppet: Not all errors are meant to propagate to the end user, especially when it is deemed rare and technical as in this case. That also helps not helping attackers. It is okay to display a generic error and log the technical error.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/freemarker/OID4VPUserAuthBean.java
IngridPuppet: I agree the caching policy was violated. Fixed as suggested.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/PresentationDuringIssuanceService.java
IngridPuppet: Thank you for raising this point. In line with the response to https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/pull/158#discussion_r3821059709, issuer_state is not OpenID4VCI specific so does not need to pass `IssuerState.fromEncodedString`. I've updated the code to avoid the uncontrolled exception as suggested.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/PresentationDuringIssuanceService.java
IngridPuppet: You've realized that the flows fundamentally behave differently in this regard, and also probably that this PR avoids changes that could break that other flow. I've opened a separate ticket to handle this task separately:
- https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/163

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/NestedPresentationDuringIssuanceTest.java
IngridPuppet: I agree, this is how many wallets work. I added a test that covers this flow: `shouldCompleteNestedAuthorizationThroughPar`. Thanks.

docs/modules/ROOT/pages/presentation-during-issuance-configuration.adoc
IngridPuppet: The challenge endpoint is already disabled when the realm flag is not set. Check AuthorizationChallengeEndpointFactory.java#L17. I've disabled redirection for the nested flow in the same direction. And added a negative test to assert the behavior. Thanks.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/presentation/GuardedCredentialScope.java
IngridPuppet: Parsing the configured list into `PresentationDuringIssuanceMode` makes logic less flexible. I've added validation on Keycloak startup. Hopefully, there is no means non-technical to update the configuration  at runtime, so this should work. I left this concern open on purpose to test the review process... Good catch! Thanks.

IngridPuppet:
Thank you @Ogenbertrand @forkimenjeckayang for your review. Your concerns should now be addressed. Please could you check again?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/OID4VPLoginActionsService.java
Ogenbertrand: Thanks, the marker is now stronger since it records both presentation mode and OpenID4VP profile. I agree the remaining transaction-level binding concern is better handled separately in #163, since changing that here could affect both nested and interactive authorization behavior.

Ogenbertrand:
I'll approve this PR once i am able to test the flow, the national wallet is breaking, it's seems active updates are been carried on there.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/presentation/GuardedCredentialScope.java
forkimenjeckayang: Startup validation is useful, but Keycloak client-scope attributes can be changed at runtime through the Admin Console or Admin REST API. `validateRealm()` currently runs only on `PostMigrationEvent`, and neither the PDI service nor the credential gate calls `validateConfiguration()`. An invalid runtime update therefore bypasses administrator-facing validation until restart.

The current gate fails closed, so I do not consider this blocking. As a follow-up, can we validate the policy when it is consumed (or in the admin update path) and add a runtime-update test for an unknown mode and a missing profile?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/PresentationVerifiedNote.java
```diff
@@ -0,0 +1,30 @@
...
+        return new PresentationVerifiedNote(mode.getValue(), profileId);
+    }
+
+    public String toJson() {
+        return JsonSerialization.valueAsString(this);
+    }
+
+    public static PresentationVerifiedNote fromJson(String json) {
+        return JsonSerialization.valueFromString(json, PresentationVerifiedNote.class);
+    }
```
forkimenjeckayang, line 29: The released `v1.3.1`–`v1.3.4` code stores this same user-session note as the literal string `"true"`. `fromJson("true")` throws `IllegalArgumentException`/`MismatchedInputException`, and the credential gate does not catch it. An active or persistent session crossing an upgrade can therefore turn a credential request into a 500.

**Supporting affected lines:** `PatchedOID4VCIssuerEndpoint.java:183-186` calls this parser without handling an invalid legacy value.

Can we parse this defensively and treat legacy or malformed values as *no valid proof*, so the gate returns its normal `invalid_credential_request` response. Do not accept the old boolean as proof because it contains no profile binding. WDYT ?

forkimenjeckayang (commented):
thank you for considering my previous comments @IngridPuppet 
One more...

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/model/PresentationVerifiedNote.java
IngridPuppet: The note is set internally and short-lived. There is no need to worry about legacy formats of the note.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/presentation/GuardedCredentialScope.java
IngridPuppet: > neither the PDI service nor the credential gate calls validateConfiguration(). An invalid runtime update therefore bypasses administrator-facing validation until restart.

I'd rather we avoid throwing exceptions just to then run after handling them the right way. The current implementation is failsafe. An unknown mode configured is ignored and a non-existing profile is logged in OID4VPUserAuthBean. The issuance gate also rejects all profiles in the latter case. One can also argue that we need to validate that the profile matches one that is configured in the authenticator. But these are all expensive validation to run on every access.

As for the admin path, it would have certainly been a great addition to do this validation on credential scope update but since we are working in a plugin, I think it is not worth the effort.

In response to your renewed concern, I've logged invalid configurations in the GuardedCredentialScope constructor to ease debugging. Please could you check?

Ogenbertrand (approved):
Thanks for considering my comments here @IngridPuppet 
I have tested and the flow completes and works for me.

IngridPuppet:
Thank you @Ogenbertrand @forkimenjeckayang for the reviews!

PR #169, opened 2026-08-31
refactor: unify credential configuration resolution

### LINKED ISSUE #163: Unify issuer-state / authorization_details / scope resolution across PresentationDuringIssuanceService and AuthorizationChallengeEndpoint
> could we extract the issuer-state / authorization_details / scope resolution into one shared, server-side resolver and make its cardinality contract explicit. The new nested resolver rejects multiple credential configurations, whereas AuthorizationChallengeEndpoint and the issuer guard select the first one or this is intended

ETA: 1 day

Awambeng:
## Problem

`PresentationDuringIssuanceService` and `AuthorizationChallengeEndpoint` both determine which credential configuration is expected from an OID4VC authorization request.

However, they handled some cases differently:

* The PDI service rejected requests with multiple credential configurations, while ACE used the first one.
* The PDI service filtered `authorization_details` by `type=openid_credential`, while ACE did not.
* Both classes contained duplicated logic for parsing the request and resolving the credential configuration.

## Solution

Introduced a shared `CredentialRequestResolver` that:

1. Uses the same resolution order: `issuer_state → authorization_details → scope`
2. Rejects requests that resolve to multiple credential configurations
3. Applies the same `openid_credential` filtering
4. Contains the shared credential offer resolution logic

Both `PresentationDuringIssuanceService` and `AuthorizationChallengeEndpoint` now use the resolver, removing the duplicated logic and ensuring consistent behavior.

Closes #163

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/CredentialRequestResolver.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/CredentialRequestResolver.java
@@ -1,0 +1,220 @@
+package io.github.adorsysgis.keycloak.protocol.oid4vc.oidc;
+
+import static org.keycloak.OID4VCConstants.OPENID_CREDENTIAL;
+import static org.keycloak.protocol.oid4vc.utils.CredentialScopeUtils.findCredentialScopeModelByConfigurationId;
+import static org.keycloak.protocol.oid4vc.utils.CredentialScopeUtils.findCredentialScopeModelByName;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import io.github.adorsysgis.keycloak.protocol.oid4vc.presentation.GuardedCredentialScope;
+import jakarta.ws.rs.BadRequestException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import org.jboss.logging.Logger;
+import org.keycloak.OAuth2Constants;
+import org.keycloak.models.ClientModel;
+import org.keycloak.models.KeycloakSession;
+import org.keycloak.models.RealmModel;
+import org.keycloak.models.oid4vci.CredentialScopeModel;
+import org.keycloak.protocol.oid4vc.issuance.credentialoffer.CredentialOfferState;
+import org.keycloak.protocol.oid4vc.issuance.credentialoffer.CredentialOfferStorage;
+import org.keycloak.protocol.oid4vc.model.IssuerState;
+import org.keycloak.protocol.oid4vc.model.OID4VCAuthorizationDetail;
+import org.keycloak.protocol.oidc.endpoints.AuthorizationEndpoint;
+import org.keycloak.sessions.AuthenticationSessionModel;
+import org.keycloak.util.JsonSerialization;
+import org.keycloak.utils.StringUtil;
+
+/**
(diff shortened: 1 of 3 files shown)


IngridPuppet:
@lightbridge-assistant review

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/presentation/AuthorizationChallengeEndpoint.java
```diff
@@ -252,8 +247,8 @@ private String resolveEnforcedProfileId(
                 return null;
             }
 
-            GuardedCredentialScope credentialScope = GuardedCredentialScope.from(
-                    resolveRequestedCredentialScope(client, scope, authorizationDetails, offerState));
+            GuardedCredentialScope credentialScope = CredentialRequestResolver.resolveCredentialScope(
```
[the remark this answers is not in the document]
IngridPuppet: This update is intentional. Maybe not the BadRequestException swallowing. Will review.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/CredentialRequestResolver.java
```diff
@@ -0,0 +1,220 @@
...
+ * authorization request. The resolution follows a strict priority order:
+ *
+ * <ol>
+ *   <li>{@code issuer_state} — the server-side credential offer bound to the issuer state,
+ *   <li>{@code authorization_details} — the JSON-encoded OID4VCI authorization details,
+ *   <li>{@code scope} — the OAuth 2.0 scope parameter (credential configuration name).
+ * </ol>
+ *
+ * <p><strong>Cardinality contract:</strong> this resolver rejects requests that reference
+ * multiple credential configurations. The OID4VCI credential endpoint issues one credential
```
IngridPuppet: It can issue multiple credentials of the same type (batch issuance).

```suggestion
 * multiple credential configurations. The OID4VCI credential endpoint processes one credential type
```

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/CredentialRequestResolver.java
```diff
@@ -0,0 +1,220 @@
...
+     * @param offerState the resolved credential offer state (from issuer_state), or {@code null}
+     * @return the resolved guarded credential scope, or {@code null} when no credential is requested
+     * @throws BadRequestException when multiple credential configurations are requested
+     */
+    public static GuardedCredentialScope resolveCredentialScope(
+            RealmModel realm,
+            ClientModel client,
+            String scope,
+            String authorizationDetails,
+            CredentialOfferState offerState) {
```
IngridPuppet: Could you pass these args in their order of precedence?

IngridPuppet (commented):
Thank you for the update. I agree with the changes. Also, all preexisting tests continue to pass. Please could you check my minor comments below?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oidc/CredentialRequestResolver.java
Awambeng: Done.

IngridPuppet (approved):

PR #174, opened 2026-09-02
Fix mDoc CBOR byte array unwrapping

### LINKED ISSUE #162: Investigate memory retention for national-wallet PID/mDoc authorization responses
## Context

Split from #133.

During national-wallet PID/mDoc presentation verification, Keycloak terminated with:

```text
Terminating due to java.lang.OutOfMemoryError: Java heap space
```

We need to identify which objects dominate memory during verification instead of only increasing heap size.

## Goal

Capture a real or representative national-wallet OpenID4VP authorization response and analyze memory usage through the verification pipeline to find the main retained objects and allocation hot spots.

## Scope

- Capture a national-wallet PID/mDoc presentation response suitable for local/reproducible analysis.
- Replay the response through the verifier path in a controlled environment.
- Generate heap dumps or allocation profiles around the verification step.
- Identify the largest retained objects and allocation sources.
- Check whether memory pressure comes from payload buffering, Base64/JSON/JWE/CBOR decoding, mDoc verification structures, certificate handling, session storage, or logging/error handling.
- Propose code-level improvements if avoidable large allocations or retained references are found.

## Acceptance criteria

- A reproducible profiling setup exists for a national-wallet PID/mDoc response.
- Heap dump or allocation profile results identify the dominant memory consumers.
- The investigation clearly separates expected mDoc verification cost from avoidable memory retention.
- Follow-up implementation tasks are created for any concrete optimization opportunities.
- Findings are documented with recommended profiling commands and environment settings.

Ogenbertrand:
closes: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/162

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/CborUtil.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/CborUtil.java
@@ -29,6 +29,13 @@ public class CborUtil {
     }
 
     /**
+     * Unwraps a field whose schema defines the byte string as embedded CBOR.
+     */
+    public static CBORItem unwrapEmbeddedCbor(CBORItem item) {
+        return unwrap(item, null, CBOR_TAG_EMBEDDED);
+    }
+
+    /**
      * Deeply unwrap CBORItem tree for convenience.
      *
      * @param item      Root of CBOR tree to unwrap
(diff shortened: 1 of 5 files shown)


src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/CborUtil.java
```diff
@@ -40,22 +47,15 @@ private static CBORItem unwrap(CBORItem item, String key, Number tagNumber) {
...
-
-                    if (decodedInnerItem == null
-                            || !numericEquals(CBOR_TAG_EMBEDDED, tagNumber)
-                                    && !(decodedInnerItem instanceof CBORTaggedItem taggedItem
-                                            && numericEquals(CBOR_TAG_EMBEDDED, taggedItem.getTagNumber()))) {
-                        yield byteArray;
-                    }
-
-                    // Continue unwrapping in case there are nested unwrapped values
-                    yield unwrap(decodedInnerItem);
```
IngridPuppet, line 58: Yes, I agree with AI on this. It's probably a good idea to leave these checks, and also the last comment on why we recurse. This method needs comments to ease understanding.

This is the real fix as per my tests:
```java
  if (!numericEquals(CBOR_TAG_EMBEDDED, tagNumber)) {
      yield byteArray;
  }
```

IngridPuppet (commented):
I confirm this enables load tests to complete. Just a minor comment.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/mdoc/MdocBaseTest.java
IngridPuppet, line 1: You should also removing this:

https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/blob/22ff42bcd4c69fa2e93665077d149acb49cd40f6/src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/KeycloakTestContainer.java#L47

IngridPuppet (approved):

PR #175, opened 2026-09-02
Monitor performance goals with load testing

### LINKED ISSUE #161: Stress test PID/mDoc presentation verification under realistic load
## Context

Split from #133.

While testing PID presentation verification with the national wallet, Keycloak terminated during authentication with:

```text
Terminating due to java.lang.OutOfMemoryError: Java heap space
```

The immediate workaround was to increase the Keycloak JVM heap, but we need a proper production-readiness investigation so this does not reappear under real traffic or larger wallet responses.

## Goal

Create a repeatable stress/load test for the OpenID4VP PID/mDoc presentation verification path and define safe resource expectations for the plugin.

## Scope

- Build a repeatable stress scenario for PID/mDoc presentation verification.
- Exercise the full Keycloak verifier path, not only isolated unit-level mDoc parsing.
- Include realistic national-wallet-sized authorization responses.
- Measure heap usage, GC behavior, request latency, and failure thresholds.
- Compare single-request large-payload behavior with concurrent verification behavior.
- Document recommended JVM/container memory settings for this flow.

## Acceptance criteria

- A reproducible test or benchmark exists for PID/mDoc presentation verification.
- The test captures memory and latency metrics under increasing concurrency/payload size.
- The failure threshold is known and documented.
- Recommended runtime settings are documented for local test and production-like deployments.
- Any discovered code-level bottlenecks are captured as follow-up issues.

**ETA:** 2 days

IngridPuppet:
Closes https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/161

Blocked by (which addresses the core heap issue with mDoc):
- #174 

Load testing revealed that no special measures should be taken because of this plugin.

A nightly workflow will monitor usability with a heap allocation as low as 512MB. Check test runs: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/actions/workflows/perf.yml

### DIFF
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -1,0 +1,91 @@
+name: Load Testing
+
+on:
+  push:
+    branches:
+      - 161-stress-testing
+  schedule:
+    # Every night at 02:00 UTC
+    - cron: '0 2 * * *'
+  workflow_dispatch:
+    inputs:
+      virtualUsers:
+        description: 'Peak number of concurrent virtual users'
+        required: false
+        default: '1000'
+      rampDurationSeconds:
+        description: 'Duration of the ramp-up phase in seconds'
+        required: false
+        default: '15'
+      steadyDurationSeconds:
+        description: 'Duration of the steady-state phase in seconds'
+        required: false
+        default: '60'
+
+permissions:
+  contents: read
+
+jobs:
+  load-test:
+    name: Run Gatling Simulations
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+
+      - name: Set up JDK 21
+        uses: actions/setup-java@v5
+        with:
+          distribution: 'temurin'
+          java-version: '21'
+
+      - name: Cache Maven packages
+        uses: actions/cache@v6
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+
+      - name: Make Maven Wrapper executable
+        run: chmod +x mvnw
+
+      - name: Build shaded plugin JAR
+        run: ./mvnw clean package -DskipTests --batch-mode
+
+      - name: Run all Gatling simulations
+        run: |
+          ./mvnw gatling:test --batch-mode \
+            -Dgatling.virtualUsers=${{ inputs.virtualUsers || '1000' }} \
+            -Dgatling.rampDurationSeconds=${{ inputs.rampDurationSeconds || '15' }} \
+            -Dgatling.steadyDurationSeconds=${{ inputs.steadyDurationSeconds || '60' }} \
+            -Dgatling.runMultipleSimulations=true
+
+      - name: Notify Discord on failure
+        if: failure()
+        run: |
+          curl -fsSL -X POST "${{ secrets.DISCORD_WEBHOOK_URL }}" \
+            -H "Content-Type: application/json" \
+            -d '{
+              "embeds": [{
+                "title": "Load testing failed",
+                "description": "See ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}",
+                "color": 15158332,
+                "fields": [
+                  {
+                    "name": "Repository",
+                    "value": "`${{ github.repository }}`",
+                    "inline": true
+                  },
+                  {
+                    "name": "Branch",
+                    "value": "`${{ github.ref_name }}`",
+                    "inline": true
+                  }
+                ],
+                "footer": {
+                  "text": "GitHub Actions"
+                }
+              }]
+            }'
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -1,9 +1,6 @@
 name: Load Testing
 
 on:
-  push:
-    branches:
-      - 161-stress-testing
   schedule:
     # Every night at 02:00 UTC
     - cron: '0 2 * * *'
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -22,6 +22,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: load-testing
+  cancel-in-progress: true
+
 jobs:
   load-test:
     name: Run Gatling Simulations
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -59,6 +59,13 @@ the IDE does not delegate the build to Maven and therefore does not produce the
 into the Keycloak container. Build the JAR explicitly (as described above) before running the tests from
 the IDE.
 
+Additionally, Gatling load tests are provided in the codebase to monitor the plugin's performance and
+behavior under heavy load. Run them with (as they are run in the CI/CD pipeline):
+
+```sh
+./mvnw gatling:test
```
Ogenbertrand, line 66: This command fails in non-interactive mode now that multiple simulations exist, please add `-Dgatling.runMultipleSimulations=true` or document how to select one simulation.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/gatling/tools/BaseLoadTest.java
```diff
@@ -0,0 +1,356 @@
...
+    /**
+     * Applies shared performance and correctness assertions to simulations: 95th percentile
+     * response time under 3s, zero failures, and at least 100 requests executed.
+     */
+    public static void assertions(Simulation.SetUp setUp) {
+        setUp.assertions(
+                global().responseTime().percentile3().lt(3000),
+                global().failedRequests().count().is(0L),
+                global().allRequests().count().gt(100L));
+    }
```
Ogenbertrand, line 134: These assertions validate latency and request success, but not heap usage. We should fail the simulation when sampled memory exceeds the expected threshold.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/gatling/tools/MemoryMonitor.java
```diff
@@ -0,0 +1,245 @@
...
+        JMXConnector jmxConnector = null;
+
+        try {
+            while (running.get()) {
+                try {
+                    if (jmxConnector == null) {
+                        jmxConnector = JMXConnectorFactory.connect(LoadTestContainer.getJmxServiceUrl(), null);
+                    }
+
+                    HeapSample sample = sampleHeap(jmxConnector.getMBeanServerConnection());
```
Ogenbertrand, line 107: The heap samples are useful, but they remain observational only unless the monitor exposes a threshold check used by the Gatling assertions.

.github/workflows/perf.yml
```diff
@@ -0,0 +1,88 @@
...
+          restore-keys: |
+            ${{ runner.os }}-maven-
+
+      - name: Make Maven Wrapper executable
+        run: chmod +x mvnw
+
+      - name: Build shaded plugin JAR
+        run: ./mvnw clean package -DskipTests --batch-mode
+
+      - name: Run all Gatling simulations
```
Ogenbertrand, line 58: You shoul upload `target/gatling/**` as an artifact with `if: always()` so reports and memory CSVs remain available after scheduled runs.

Ogenbertrand (changes requested):
Great i ran the test locally and it looks perfect, i have just a few comments could you check?

Ran a heavier test closer to CI defaults:
```
./mvnw gatling:test --batch-mode \
  -Dgatling.virtualUsers=1000 \
  -Dgatling.rampDurationSeconds=15 \
  -Dgatling.steadyDurationSeconds=60 \
  -Dgatling.runMultipleSimulations=true
```

Image

And a quick live smoke test:
```
./mvnw gatling:test --batch-mode \
  -Dgatling.virtualUsers=2 \
  -Dgatling.rampDurationSeconds=1 \
  -Dgatling.steadyDurationSeconds=5 \
  -Dgatling.runMultipleSimulations=true
```

Image

Image

README.md
IngridPuppet: Thank you. The command does not fail, it prompts for available simulations; unless I'm missing something. I'll refer to the Gatling documentation for full options.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/gatling/tools/BaseLoadTest.java
IngridPuppet: Heap monitoring does not natively integrate with Gatling. The heap memory on the test container is capped at 512MB. A heap overflow will cause the test to fail as a consequence.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/gatling/tools/MemoryMonitor.java
IngridPuppet: Heap monitoring does not natively integrate with Gatling. The heap memory on the test container is capped at 512MB. A heap overflow will cause the test to fail as a consequence.

.github/workflows/perf.yml
IngridPuppet: Done. See https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/actions/runs/33765098974

README.md
Ogenbertrand: Good point, it prompts interactively. Since this section says it is the CI/CD command, could we still show the non-interactive CI equivalent with `-Dgatling.runMultipleSimulations=true` and keep the Gatling docs link for the other options?

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/gatling/tools/BaseLoadTest.java
Ogenbertrand: That makes sense. Since the container heap is capped, an OOM will fail the simulation, and the CSV output is enough as diagnostic evidence for this PR.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/gatling/tools/MemoryMonitor.java
Ogenbertrand: Agreed, with the 512MB container cap the monitor can remain diagnostic-only because an overflow fails the test run.

.github/workflows/perf.yml
Ogenbertrand: Confirmed, this addresses the report retention comment. Thanks.

.github/workflows/perf.yml
```diff
@@ -0,0 +1,88 @@
...
+        description: 'Duration of the steady-state phase in seconds'
+        required: false
+        default: '60'
+
+jobs:
+  load-test:
+    name: Run Gatling Simulations
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
```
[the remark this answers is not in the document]
Ogenbertrand: Resolved by adding workflow-level `permissions: contents: read`.

README.md
IngridPuppet: > Since this section says it is the CI/CD command

It says that within parentheses, meaning not necessarily verbatim. The simple command is the best default for humans who are going to read the README. They are prompted properly.

Ogenbertrand (approved):
LGTM

IngridPuppet:
Thank you @Ogenbertrand for the review. We'll watch out nightly runs and see how to improve the setup.

PR #181, opened 2026-09-03
feat: improve claim binding comparison

### LINKED ISSUE #179: Enforce claim binding with some flexibility
Although binding comparators can be plugged in, the default logic is too rigid:
- Binding can fail because of case sensitiveness ("JOHN" vs "John")
- Binding can fail because of trailing spaces (" John  " vs "John")

For the former, add an attribute to enable case-insentive comparisons.
For the latter, trail spaces systematically.

ETA: 1 day

Awambeng:
This PR improves the claim binding comparison by trimming whitespace and supporting optional case-insensitive matching.

## Changes

* Added `caseInsensitive` to `BindingRule`
* Trimmed leading and trailing whitespace before comparing claim values
* Added optional case-insensitive matching
* Kept case-sensitive matching as the default for backward compatibility
* Added tests for trimming, case-insensitive matching, and both together

## Configuration Example

```json
{
    "id": "my-profile",
    "credentials": [
        {
            "id": "identity",
            "role": "identity",
            "credentialTypes": [
                "identity-vct"
            ],
            "claims": [
                "family_name"
            ],
            "binding": [
                {
                    "type": "claim_equals_user_attribute",
                    "credentialClaim": "family_name",
                    "userAttribute": "lastName",
                    "caseInsensitive": true // Ignore case when comparing
                }
            ]
        }
    ]
}
```

Closes #179

### DIFF
--- a/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
+++ b/src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/oid4vp/authenticator/OID4VPAuthenticator.java
@@ -20,6 +20,7 @@ import jakarta.ws.rs.core.Response;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.LinkedHashMap;
+import java.util.Locale;
 import java.util.Map;
 import org.jboss.logging.Logger;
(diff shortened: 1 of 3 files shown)


IngridPuppet (approved):
LGTM

PR #199, opened 2026-09-09
ci: skip load tests when the commit is unchanged

### LINKED ISSUE #196: Configure load testing workflow to only run if commit changed since last run
There is no need to run the workflow if nothing changed since last time it ran successfully.

ETA: 2 hours

forkimenjeckayang:
## Summary

The scheduled load-testing workflow was running Gatling repeatedly even when the repository commit had not changed since the previous successful run.

This change adds a lightweight check that:

- retrieves the latest successful load-testing workflow run for the current branch
- compares its commit SHA with the current commit
- skips the expensive build and Gatling jobs when the commit was already tested successfully
- always permits manually dispatched load tests
- reruns tests when a new commit has not succeeded yet
- runs the tests as a safe fallback if the workflow history cannot be retrieved
- respects workflow cancellation

The workflow also records its decision and reason in the GitHub Actions job summary.

Closes #196

### DIFF
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -20,6 +20,7 @@ on:
         default: '60'
 
 permissions:
+  actions: read
   contents: read
 
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/perf.yml
```diff
@@ -20,15 +20,88 @@ on:
...
+              return;
+            }
+
+            try {
+              const branch = context.ref.replace(/^refs\/heads\//, '');
+              const { data } = await github.rest.actions.listWorkflowRuns({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                workflow_id: 'perf.yml',
+                branch,
```
Ogenbertrand, line 70: Manual runs can use different load parameters, but this query includes successful workflow_dispatch runs. A manual run on the same SHA could therefore suppress the next scheduled run. We should filter by event: 'schedule' or include the load parameters in the deduplication check.

Ogenbertrand (commented):
Thanks for the PR @forkimenjeckayang 
I have one small comment

.github/workflows/perf.yml
forkimenjeckayang: updated the workflow-runs query to filter by `event: 'schedule'`. Manual dispatches will still always run, but only successful scheduled runs are now considered when deciding whether the current commit has already been tested.

IngridPuppet (commented):
Hello @forkimenjeckayang - Thank you for the PR. I reviewed the changes and have no concerns. However, I don't see any trial runs of the workflow: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/actions/workflows/perf.yml. Do you just trust your updates?

Ogenbertrand (approved):
Thanks for considering my comment.

forkimenjeckayang:
> However, I don't see any trial runs of the workflow: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/actions/workflows/perf.yml. Do you just trust your updates?

I ran the workflow manually against the PR branch which completed successfully: https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/actions/runs/34451324583

The new check job completed successfully, recognized the manual dispatch, produced `should-run=true`, and started the Gatling job. The complete build, Gatling simulations, and report upload also passed.

The scheduled skip path cannot be triggered from the PR branch because scheduled workflows execute from the default branch, while manual runs intentionally bypass deduplication. After merge, the first scheduled run should execute for the new commit, and the next scheduled run should skip if `main` remains unchanged.

IngridPuppet (approved):
LGTM

PR #201, opened 2026-09-09
Hide empty "Or sign in with" when OpenID4VP is not configured

### LINKED ISSUE #200: Hide empty "Or sign in with" when OpenID4VP is not configured
**Description**

The `keycloak.v2+oid4vp` login page always shows **"Or sign in with"** even when there are no wallet buttons and no social IdPs. That happens because `OID4VPLoginFormsProvider` always injects the `oid4vp` bean, and `login-default.ftl` treats `oid4vp??` as “show the social section.” `social-providers.ftl` then prints the heading over an empty list.

Image

EST: 1day

Vitalisn4:
### Summary
- The `keycloak.v2+oid4vp` login page showed **Or sign in with** with no buttons when OID4VP was not configured, because the theme treated the always-injected `oid4vp` bean as a reason to render the social footer.

### Test plan
- [ ] Login with OID4VP unconfigured and no social IdPs: username/password only (no "Or sign in with")
- [ ] Login with a wallet profile configured: heading + wallet CTA shown

Screenshot from 2026-09-09 13-37-40

Closes https://github.com/ADORSYS-GIS/keycloak-oid4vp-plugin/issues/200

### DIFF
--- a/src/main/resources/theme/keycloak.v2+oid4vp/login/social-providers.ftl
+++ b/src/main/resources/theme/keycloak.v2+oid4vp/login/social-providers.ftl
@@ -1,7 +1,8 @@
 <#macro show social>
-    <#assign socialProviders = (social.providers)![]>
-    <#assign oid4vpProviders = []>
-    <#if oid4vp?? && oid4vp.loginProfiles?? && oid4vp.loginProfiles?has_content>
+<div id = "kc-social-providers" class="${properties.kcFormSocialAccountSectionClass!}">
+
+    <#if oid4vp?? && oid4vp.loginProfiles?? && (oid4vp.loginProfiles?size > 0)>
+        <#assign oid4vpProviders = []>
         <#list oid4vp.loginProfiles as profile>
             <#assign oid4vpProviders = oid4vpProviders + [{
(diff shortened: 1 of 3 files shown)


src/main/resources/theme/keycloak.v2+oid4vp/login/social-providers.ftl
IngridPuppet, line 1: Is any change to this file actually necessary?

Ogenbertrand (dismissed):
Thanks for the PR @Vitalisn4 
I found no functional issues. we can merge unless @IngridPuppet  has capacity to check.

src/main/resources/theme/keycloak.v2+oid4vp/login/social-providers.ftl
Vitalisn4: Thank you, just notice changing this file isn’t needed for the fix. I added an extra check here so it wouldn’t print “Or sign in with” if the button list was empty, but that was unnecessary. `login-default.ftl` already skips this template when there’s nothing to show, so it never runs when OID4VP isn’t configured. will revert the file.

IngridPuppet (approved):
LGTM

PR #207, opened 2026-09-16
Fix x5c chain validation against trusted anchors

IngridPuppet:
So the trust chain of the test PID can be validated.

See test suite added.


IngridPuppet:
@lightbridge-assistant review

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtilTest.java
```diff
@@ -0,0 +1,285 @@
...
+
+    @BeforeAll
+    static void setup() throws Exception {
+        CryptoIntegration.init(PKIXVerificationUtilTest.class.getClassLoader());
+        List<X509Certificate> chain = readPemChain("/tokenstatus/de-pid-provider-test.pem");
+        leaf = chain.get(0);
+        intermediate = chain.get(1);
+        root = chain.get(2);
+
+        // An unrelated certificate for testing scenarios
```
Ogenbertrand, line 70: i feel that comment is uneccessary since the variable states it `unrelatedKeyPair`

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtilTest.java
Ogenbertrand, line 1: I'd add dedicated tests for expired and possibly not-yet-valid presented certificates.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtilTest.java
```diff
@@ -0,0 +1,285 @@
...
+            VerificationException error = assertThrows(
+                    VerificationException.class,
+                    () -> PKIXVerificationUtil.validateChain(
+                            testCase.chain(), testCase.roots(), testCase.intermediates()));
+            assertEquals(testCase.expectedMessage(), error.getMessage());
+        }
+    }
+
+    @SuppressWarnings("SameParameterValue")
+    private static List<X509Certificate> readPemChain(String resource) throws Exception {
```
Ogenbertrand: can we removes the warning instead of suppressing it?
```suggestion
private static List<X509Certificate> readTestChain() throws Exception {
    try (InputStream stream =
            PKIXVerificationUtilTest.class.getResourceAsStream(
                    "/tokenstatus/de-pid-provider-test.pem")) {
        ...
    }
}
```

Ogenbertrand (commented):
some little comments

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtil.java
```diff
@@ -56,7 +58,7 @@ public static X509Certificate[] validateChain(List<X509Certificate> certs, Trust
         return validateChain(certs, roots, intermediates);
     }
 
-    private static X509Certificate[] validateChain(
+    public static X509Certificate[] validateChain(
```
Awambeng: Do we need to make this public? The only usage outside the class seems to be the new test. Since the test is already in the same package, could we keep this package-private and avoid adding it to the public API?

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtil.java
```diff
@@ -75,67 +77,83 @@ private static X509Certificate[] validateChain(
...
+            cert.checkValidity(new Date(Time.currentTimeMillis()));
+            if (i == certs.size() - 1) {
+                break;
+            }
+
+            X509Certificate next = certs.get(i + 1);
+            if (!cert.getIssuerX500Principal().equals(next.getSubjectX500Principal())) {
+                throw new VerificationException("Certificate chain is not in order");
+            }
+
```
Awambeng, line 147: Is it intentional to require clients to present a contiguous chain? A client could exclude the intermediate, which could still result in a valid chain if the intermediate is available in the truststore, but the validation would fail because of this strict requirement. Is that what we want? If yes, maybe it would be good to add a test for this explicitly.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtil.java
```diff
@@ -75,67 +77,83 @@ private static X509Certificate[] validateChain(
...
+     * are consequently validated here by this weaker property (validity, issuer DN,
+     * signature) only - before the trust anchor PKIX performs its full validation.
+     */
+    private static void validatePresentedChain(List<X509Certificate> certs)
+            throws VerificationException, GeneralSecurityException {
+        for (int i = 0; i < certs.size(); i++) {
+            X509Certificate cert = certs.get(i);
+            // Fail fast on expired certificates so callers observe a CertificateExpiredException
+            // instead of the path builder's generic "unable to find valid certification path".
+            cert.checkValidity(new Date(Time.currentTimeMillis()));
```
Awambeng: I think saying callers see `CertificateExpiredException` is a bit misleading. The exception is wrapped below, so they actually get `Certificate chain validation failed`, with the expiry exception as the cause. Maybe we can phrase this around the expiry being preserved as the cause instead of the generic path validation error. WDYT?

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtilTest.java
```diff
@@ -0,0 +1,285 @@
...
+ *
+ * <p>Trust certificates can be passed as root or as intermediate. A certificate passed as a
+ * root is trusted as a root - self-signed or not - and root trust prevails if the same
+ * certificate is also listed as an intermediate. The presented chain is only valid when it
+ * chains up to a trusted root, using intermediates to bridge missing links. Trusting no
+ * root at all is always invalid.
+ */
+class PKIXVerificationUtilTest {
+
+    private static final int OCT_2026_EPOCH_SECONDS = 1790812800;
```
Awambeng, line 44: should we add a short comment explaining what date `OCT_2026_EPOCH_SECONDS` represents and that it needs to stay within the validity period of the certificates in the PEM fixture. This would make it easier to keep the test in sync if the certs are regenerated later.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtilTest.java
```diff
@@ -0,0 +1,285 @@
...
+
+        if (testCase.expectedMessage() == null) {
+            X509Certificate[] validated = assertDoesNotThrow(() ->
+                    PKIXVerificationUtil.validateChain(testCase.chain(), testCase.roots(), testCase.intermediates()));
+            assertArrayEquals(testCase.chain().toArray(new X509Certificate[0]), validated);
+        } else {
+            VerificationException error = assertThrows(
+                    VerificationException.class,
+                    () -> PKIXVerificationUtil.validateChain(
+                            testCase.chain(), testCase.roots(), testCase.intermediates()));
```
Awambeng, line 322: `validateChain(...)` is repeated in both branches. Could we pull it out so the branches only handle the different expectations? Would make the test a bit cleaner.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtilTest.java
```diff
@@ -0,0 +1,285 @@
...
+        return new Date((OCT_2026_EPOCH_SECONDS + dayOffset * 86400) * 1000);
+    }
+
+    private static List<X509Certificate> chain() {
+        return list(leaf, intermediate, root);
+    }
+
+    private static List<X509Certificate> list(X509Certificate... certificates) {
+        return List.of(certificates);
+    }
```
Awambeng: Some cases use `chain()` while others use `list(leaf, intermediate, root)` for the same thing. Maybe we can stick to one style so the test cases are consistent.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/TestCryptoUtils.java
```diff
@@ -52,15 +54,31 @@ public static X509Certificate createLeafCert(KeyPair kp, KeyPair caKp, X509Certi
...
+            X509Certificate caCert,
+            String subject,
+            boolean isCa,
+            Integer keyUsageBitmask,
+            Date notBefore,
+            Date notAfter)
+            throws Exception {
+        // The names are taken from principals directly so the encoded attribute order is
+        // preserved. A string round trip (new X500Name(principal.getName())) reverses
+        // multi-RDN names and breaks byte-level DN equality with the issuer certificate.
```
Awambeng: I think making this as a javadoc on the `createLeafCert` method will be good for callers

Awambeng (changes requested):
Thank you for the fix and the solid test cases @IngridPuppet . I left a few minor comments.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtilTest.java
IngridPuppet: It helps to visually organize the instructions in the method into sections. I'd rather keep it.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtilTest.java
IngridPuppet: Thanks, added.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtilTest.java
IngridPuppet: In such cases, especially in tests, I usually prefer to suppress the warning because we may soon need to pass another value. But applied the suggestion this time. Thanks.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtil.java
IngridPuppet: I made it package-private. Thanks.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtil.java
IngridPuppet: Yes, this is intentional. It is best to enforce a contiguous chain. Many tests covered the enforcement of contiguous chains, but I added one more specific test case to cover this scenario.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtilTest.java
IngridPuppet: I agree. Done.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtilTest.java
IngridPuppet: It is just one instruction, that's fine IMO. One branch should throw, the other not. Pulling it out makes things worse.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtilTest.java
IngridPuppet: Yes, I forgot to fix this. Removed chain() to always be explicit.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/TestCryptoUtils.java
IngridPuppet: Done.

src/main/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtil.java
IngridPuppet: I found the comment altogether misleading because the check is actually needed, not just for improving error handling. So I removed it. Thanks.

src/test/java/io/github/adorsysgis/keycloak/protocol/oid4vc/crypto/PKIXVerificationUtilTest.java
Awambeng: Thanks

Awambeng (approved):

Ogenbertrand (commented):
Thanks LGTM

Ogenbertrand (approved):

IngridPuppet:
Thank you for the reviews!

